PR #1, opened 2020-04-13
Initital commit.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

[x] Add `Form::class`.
[x] Add `FormInterface::class`.


src/Form.php
```diff
@@ -0,0 +1,327 @@
...
+    {
+        return $attribute === null ? !empty($this->errors) : isset($this->errors[$attribute]);
+    }
+
+    /**
+     * Populates the model with input data.
+     *
+     * This method provides a convenient shortcut for:
+     *
+     * ```php
```
rustamwin, line 239: ?

terabytesoftw: done.

tests/Stubs/LoginForm.php
```diff
@@ -0,0 +1,66 @@
...
+    public function getAttributesLabels(): array
+    {
+        return [
+            'login' => 'Login:',
+            'password' => 'Password:',
+            'rememberMe' => 'remember Me:'
+        ];
+    }
+
+    public function getFormname(): ?string
```
xepozz, line 62: I think `form name` can't be null...
```suggestion
    public function getFormName(): ?string
```

src/FormInterface.php
```diff
@@ -0,0 +1,24 @@
...
+    public function clearErrors(?string $attribute = null): void;
+    public function getAttributes(): array;
+    public function getAttributeHint(string $attribute): string;
+    public function getAttributesHints(): array;
+    public function getAttributeLabel(string $attribute): string;
+    public function getAttributesLabels(): array;
+    public function getErrors(?string $attribute = null): array;
+    public function getFirstError(string $attribute): ?string;
+    public function getFirstErrors(): array;
+    public function getFormname(): ?string;
```
xepozz, line 22: I think `form name` can't be null...
```suggestion
    public function getFormName(): ?string;
```

src/FormInterface.php
```diff
@@ -0,0 +1,24 @@
...
+    public function attributesLabels(): array;
+    public function clearErrors(?string $attribute = null): void;
+    public function getAttributes(): array;
+    public function getAttributeHint(string $attribute): string;
+    public function getAttributesHints(): array;
+    public function getAttributeLabel(string $attribute): string;
+    public function getAttributesLabels(): array;
+    public function getErrors(?string $attribute = null): array;
+    public function getFirstError(string $attribute): ?string;
+    public function getFirstErrors(): array;
```
xepozz, line 21: What is the "first errors"?

src/FormInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form;
+
+interface FormInterface
+{
+    public function addError(string $attribute, string $error): void;
```
xepozz, line 9: I don't sure that this method should be in the `FormInterface` and have `public` modifier

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * Sets the attribute values in a massive way.
+     *
+     * @param array $values attribute values (name => value) to be assigned to the model.
+
+     * {@see safeAttributes()}
+     * {@see attributes()}
+     */
+    public function setAttributes($values): void
+    {
+        if (\is_array($values)) {
```
xepozz, line 341: You should type the parameters `$values` as `array` instead

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * `$data`,
+     * instead of `$data['FormName']`.
+     *
+     * Note, that the data being populated is subject to the safety check by [[setAttributes()]].
+     *
+     * @param array $data the data array to load, typically `$_POST` or `$_GET`.
+     *
+     * @return bool whether `load()` found the expected form in `$data`.
+     */
+    public function load($data): bool
```
xepozz, line 314: How about to add this method to the interface?

xepozz, line 314:
```suggestion
    public function load(array $data): bool
```

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * Returns the first error of the specified attribute.
+     *
+     * @param string $attribute attribute name.
+     *
+     * @return string|null the error message. Null is returned if no error.
+     *
+     * {@see getErrors()}
+     * {@see getFirstErrors()}
+     */
+    public function getFirstError(string $attribute): string
```
xepozz, line 246: Maybe we need to return `null` when errors are absent?

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * ]
+     * ```
+     *
+     * {@see getFirstErrors()}
+     * {@see getFirstError()}
+     */
+    public function getErrors(?string $attribute = null): array
+    {
+        if ($attribute === null) {
+            return $this->errors === null ? [] : $this->errors;
```
xepozz, line 230: It can't be null by definition

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        return isset($this->errors[$attribute]) ? $this->errors[$attribute] : [];
```
xepozz, line 233:
```suggestion
        return $this->errors[$attribute] ?? [];
```

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * By default, this method returns the model class name (without the namespace part) as the form name. You may
+     * override it when the model is used in different forms.
+     *
+     * @return string|null the form name of this model class.
+     *
+     * {@see load()}
+     */
+    public function getFormname(): ?string
+    {
+        return null;
```
xepozz, line 199: Would be nice to name form by default as a classname like in Yii2

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     *
+     * @return string the attribute hint.
+     *
+     * {@see getAttributesHints()}
+     */
+    public function getAttributeHint(string $attribute): string
+    {
+        $hints = $this->getAttributesHints();
+
+        return isset($hints[$attribute]) ? $hints[$attribute] : '';
```
xepozz, line 139:
```suggestion
        return $hints[$attribute] ?? '';
```

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * @return string the attribute label.
+     *
+     * {@see generateAttributeLabel()}
+     * {@see attributesLabels()}
+     */
+    public function getAttributeLabel(string $attribute): string
+    {
+        $this->attributesLabels = $this->getAttributesLabels();
+
+        return isset($this->attributesLabels[$attribute])
```
xepozz, line 175: Use short syntax `??` instead

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * @param string $attribute the attribute name.
+     *
+     * @return string the attribute label.
+     *
+     * {@see generateAttributeLabel()}
+     * {@see attributesLabels()}
+     */
+    public function getAttributeLabel(string $attribute): string
+    {
+        $this->attributesLabels = $this->getAttributesLabels();
```
xepozz, line 173: Why you need to update a local property?

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+     * @param array $items a list of errors. The array keys must be attribute names.
+     * The array values should be error messages. If an attribute has multiple errors, these errors must be given in
+     * terms of an array.
+     *
+     * You may use the result of {@see getErrors()} as the value for this parameter.
+     */
+    public function addErrors(array $items): void
+    {
+        foreach ($items as $attribute => $errors) {
+            if (\is_array($errors)) {
```
xepozz, line 43: You can cast `$errors` to `array` instead of use conditions, i.e. 
```php
$val = 'Some text';
$vals = ($array)$val; // $vals = ['Some text'];
```
And your code may be short:
```php
foreach ((array)$errors as $error) {
   $this->addError($attribute, $error);
}
```

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+
+namespace Yiisoft\Form;
+
+use Yiisoft\Strings\Inflector;
+
+abstract class Form implements FormInterface
+{
+    private array $attributes = [];
+    private array $attributesLabels = [];
+    private array $errors = [];
```
xepozz, line 13:
```suggestion
    private array $attributesErrors = [];
```

tests/TestCase.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
xepozz, line 1: I think you no need this file. Please check the LoginForm directly in test class

src/FormInterface.php
```diff
@@ -0,0 +1,24 @@
...
+    public function addErrors(array $items): void;
+    public function attributes(): array;
+    public function attributesLabels(): array;
+    public function clearErrors(?string $attribute = null): void;
+    public function getAttributes(): array;
```
xepozz, line 14: What kind of differences of this method and `attributes`?

src/Form.php
```diff
@@ -0,0 +1,378 @@
...
+    /**
+     * Returns the list of attribute names.
+     *
+     * By default, this method returns all private non-static properties of the class.
+     *
+     * @return array list of attribute names.
+     */
+    public function attributes(): array
+    {
+        $class = new \ReflectionClass($this);
```
xepozz, line 62: I think you need to do this in constructor instead of public method

tests/Stubs/LoginForm.php
terabytesoftw: done.

src/FormInterface.php
terabytesoftw: done.

src/FormInterface.php
terabytesoftw: We can have a field that has several validations and several error messages, this function returns the first error.

src/FormInterface.php
terabytesoftw: This method allows addError in the validation, it must be public.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done, separate methods `getError(), getErrors().`

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/FormInterface.php
terabytesoftw: attributes retrieve propertys with typehints.

tests/TestCase.php
terabytesoftw: I will do it at the end if no more classes are added.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
```diff
@@ -0,0 +1,373 @@
...
+
+        $type = null;
+
+        foreach ($class->getProperties(\ReflectionProperty::IS_PRIVATE) as $property) {
+            if (!$property->isStatic()) {
+                $type = (new \ReflectionProperty($property->class, $property->name))->getType();
+
+                if ($type !== null) {
+                    $this->attributes[$property->getName()] = $type->getName();
```
xepozz, line 345: What about to store data in  the following structure:
```php
attributes: [
 {"type": "bool", "value": true, "modifier": "private", "setter": "setMyBestValue", etc...},
 {"type": "object", "value": *object*, "modifier": "public", "setter": null, etc...},
]
```

src/Form.php
```diff
@@ -0,0 +1,373 @@
...
+     *
+     * For example, 'department_name' or 'DepartmentName' will generate 'Department Name'.
+     *
+     * @param string $name the column name.
+     *
+     * @return string the attribute label.
+     */
+    private function generateAttributeLabel(string $name): string
+    {
+        $inflector = new Inflector();
```
xepozz, line 369: I think we need to store object in the private property for cache reasons

tests/bootstrap.php
```diff
@@ -0,0 +1,14 @@
+<?php
```
xepozz, line 1: It's not necessary. You can just add one line to config
https://github.com/yiisoft/http/blob/master/phpunit.xml.dist#L6

src/FormInterface.php
```diff
@@ -0,0 +1,24 @@
...
+    public function getAttributeHint(string $attribute): string;
+    public function getAttributesHints(): array;
+    public function getAttributeLabel(string $attribute): string;
+    public function getError(string $attribute): ?array;
+    public function getErrors(): ?array;
+    public function getFirstError(string $attribute): ?string;
+    public function getFirstErrors(): array;
+    public function formname(): string;
```
xepozz, line 21:
```suggestion
    public function formName(): string;
```

src/Form.php
```diff
@@ -0,0 +1,373 @@
...
+     *
+     * @return bool whether `load()` found the expected form in `$data`.
+     */
+    public function load(array $data): bool
+    {
+        $result = false;
+
+        $scope = $this->formName();
+
+        if ($scope === null && !empty($data)) {
```
xepozz, line 287: `$scope` can't be `null`

src/Form.php
samdark: Why?

PR #3, opened 2020-04-14
Add support yiisoft/validators.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

[x] Add support yiisoft/validators.
[x] Implement DataSetInterface.
[x] Add tests all cases.
[x] Update README.md.
[x] Remove travis.yml.
[x] Add support `propertyReader()`
[x] Add support `propertyWriter()`


README.md
```diff
@@ -92,6 +95,28 @@ class LoginForm extends Form
...
+
+    /** Define login rules */
+    private function loginRules(): array
+    {
+        return [
+            new Required(),
+            (new HasLength())
+            ->min(4)
+            ->max(40)
+            ->tooShortMessage('Is to short.')
```
schmunk42, line 115: should read `'Is too short.'`, also below and in tests

terabytesoftw: done.

src/Form.php
```diff
@@ -80,6 +84,17 @@ public function clearErrors(?string $attribute = null): void
         }
     }
 
+    public function getAttributeValue(string $attribute)
```
schmunk42, line 87: Since we're using magic getters, was there ever a consideration **not** to use class methods starting with `get...`? 
I.e. also `getError()` but rather something like `resolveError()` or `retrieveError()`, might be late to the party here, but I stumbled upon that in Yii2 quite sometimes.

CC: @samdark

terabytesoftw:
ok no problem i will add a suffix parameter to the methods so there will be no problem using any name

src/Form.php
terabytesoftw: ok no problem i will add a suffix parameter to the methods so there will be no problem using any name

src/Form.php
```diff
@@ -323,6 +343,25 @@ public function setAttributes(array $values): void
...
+    public function validate(): bool
+    {
+        $this->clearErrors();
+
+        if (!empty($this->attributesRules)) {
+            $results = (new Validator($this->attributesRules))->validate($this);
+
+            foreach ($results as $attribute => $result) {
+                if ($result->isValid() === false) {
+                    foreach ($result->getErrors() as $error) {
```
xepozz, line 355: `$this->addErrors([$attribute => $result->getErrors()]);`

src/Form.php
```diff
@@ -5,18 +5,22 @@
...
     private array $attributesLabels;
     private array $attributesErrors = [];
+    private array $attributesRules = [];
     private ?Inflector $inflector = null;
 
     public function __construct()
     {
         $this->attributes = $this->attributes();
         $this->attributesLabels = $this->attributeLabels();
+        $this->attributesRules = $this->rules();
```
xepozz, line 23: I think you need to keep empty `rules` into this class, because form can exist without any rules

src/Form.php
```diff
@@ -171,7 +191,7 @@ public function formName(): string
      */
     public function getError(string $attribute): ?array
```
xepozz, line 192: why result type is `?array`?

src/Form.php
terabytesoftw: Done.

src/Form.php
terabytesoftw: Done.

src/Form.php
terabytesoftw: Done.

src/Form.php
terabytesoftw: Done,

src/Form.php
```diff
@@ -162,6 +177,11 @@ public function formName(): string
         return '';
     }
 
+    public function hasAttribute(string $attribute): bool
+    {
+        return (new \ReflectionClass($this))->hasProperty($attribute);
```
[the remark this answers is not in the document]
terabytesoftw: Done.

src/Form.php
```diff
@@ -371,4 +443,26 @@ private function generateAttributeLabel(string $name): string
 
         return $this->inflector->camel2words($name, true);
     }
+
+    private function propertyReader(string $attribute)
+    {
+        $getter = fn ($class, $attribute) => $class->$attribute ?? null;
```
xepozz, line 449: `$value = $this->$attribute`?

terabytesoftw: Closure#bind() basically allows you to get an instance of a closure basically like adding APIs to existing objects!

xepozz: Yeah, but why we need it?

xepozz: You should:
1. Get value from `$this->get{attribute}()` method if it exist
2. Or get value from `$this->{attribute}` if it exist
3. Or throw error

terabytesoftw: Directly getValue() of the properties for getAttributes($attribute), it allows to use any name for the methods, for example if you have getError(), isError(), resolveError(), how are you going to get the method name for the property, that's why only for getAttributeValue () makes sense to directly access the property.

https://github.com/terabytesoftw/yii-form/blob/a1b5ac32ebbc48952d35c4b62a4e409f89c5cac7/src/Form.php#L455

xepozz:
```php
class MyForm
{
  private string $time;
  public function getTime(): DateTimeInterface { ... }

  public function rules():
  {
    return ['time' => (new DateTimeValidator())->notToday()];
  }
}
```
What kind of value will we  have in the `DateTimeValidator` validator or others?

terabytesoftw: the validator takes the values of getAttributesValue from DataSetInterface review tests.

src/Form.php
xepozz: I think it's rarely case. Getters and setters are exist in all the world. If you want to use non-standart getter method name, you can just call the getter in your custom method.

src/Form.php
terabytesoftw: Another thing to think about that form will interact with widgets like `ActiveForm::class`, `ActiveField::class`, `AR`, `ORM` and custom widgets, where each developer has different ways of working, so working with the properties directly will be the best we will avoid many bugs , or more complex code to guess the names.

src/Form.php
```diff
@@ -371,4 +443,22 @@ private function generateAttributeLabel(string $name): string
 
         return $this->inflector->camel2words($name, true);
     }
+
+    private function propertyReader(string $attribute)
```
samdark, line 447:
```suggestion
    private function readProperty(string $attribute)
```

src/Form.php
```diff
@@ -371,4 +443,22 @@ private function generateAttributeLabel(string $name): string
...
+    {
+        $getter = fn ($class, $attribute) => $class->$attribute;
+        $getter = \Closure::bind($getter, null, $this);
+
+        $result = $getter($this, $attribute);
+
+        return $result;
+    }
+
+    private function propertyWriter(string $attribute, $value): void
```
samdark, line 457:
```suggestion
    private function writeProperty(string $attribute, $value): void
```

README.md
```diff
@@ -92,6 +95,28 @@ class LoginForm extends Form
     {
         return 'LoginForm';
     }
+
+    /** Add rules */
+    public function rules(): array
```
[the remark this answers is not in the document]
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

src/Form.php
terabytesoftw: done.

samdark:
Thank you!

PR #4, opened 2020-04-16
Add getErrorSummary().

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

[x] `Add getErrorSummary()` need for Html Forms.

### DIFF
--- a/tests/Stubs/LoginForm.php
+++ b/tests/Stubs/LoginForm.php
@@ -70,7 +70,8 @@ class LoginForm extends Form
     protected function rules(): array
     {
         return [
-            'login' => $this->loginRules()
+            'login' => $this->loginRules(),
+            'password' => $this->passwordRules()
         ];
     }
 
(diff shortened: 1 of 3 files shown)


tests/Stubs/LoginForm.php
```diff
@@ -86,4 +87,14 @@ private function loginRules(): array
             new Email()
         ];
     }
+
+    private function passwordRules(): array
+    {
+        return [
+            new Required(),
+            (new HasLength())
```
samdark, line 95: This rule should be removed. It's a bad practice to limit password length.

terabytesoftw:
```
        return [
            new Required(),
            (new HasLength())
            ->min(8)
            ->tooShortMessage('Is too short.'),
        ];
```

terabytesoftw: done.

samdark:
👍

PR #15, opened 2020-05-15
Add method for adding form errors manually.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

Adds an ability to add errors to the form manually. The use case is when you do some external validation (that is not in rules) such as validating password with database value (I prefer not to place db dependencies in `FormModel` and do validation manually in some service).

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -215,6 +215,11 @@ abstract class FormModel implements FormModelInterface
         return !$this->hasErrors();
     }
 
+    public function addError(string $attribute, string $error): void
+    {
+        $this->attributesErrors[$attribute][] = $error;
+    }
+
     private function addErrors(array $items): void
     {
         foreach ($items as $attribute => $errors) {
--- a/src/FormModelInterface.php
+++ b/src/FormModelInterface.php
@@ -78,6 +78,14 @@ interface FormModelInterface extends DataSetInterface
     public function isAttributeRequired(string $attribute): bool;
 
     /**
+     * Add error for the specified attribute.
+     *
+     * @param string $attribute attribute name.
+     * @param string $error attribute error message.
+     */
+    public function addError(string $attribute, string $error): void;
+
+    /**
      * Returns the errors for all attributes.
      *
      * @return array errors for all attributes or the specified attribute. null is returned if no error.
--- a/tests/FormModelTest.php
+++ b/tests/FormModelTest.php
@@ -7,6 +7,8 @@ namespace Yiisoft\Yii\Form\Tests;
 use InvalidArgumentException;
 use Yiisoft\Yii\Form\Tests\Stub\LoginForm;
 
+use function str_repeat;
+
 final class FormModelTest extends TestCase
 {
     public function testGetFormName(): void
@@ -171,7 +173,7 @@ final class FormModelTest extends TestCase
             $form->error('login')
         );
 
-        $form->login(\str_repeat('x', 60));
+        $form->login(str_repeat('x', 60));
         $form->validate();
         $this->assertEquals(
             'Is too long.',
--- a/tests/FormModelTest.php
+++ b/tests/FormModelTest.php
@@ -130,6 +130,17 @@ final class FormModelTest extends TestCase
         $this->assertEquals(true, $form->getRememberMe());
     }
 
+    public function testAddError(): void
+    {
+        $form = new LoginForm();
+        $errorMessage = 'Invalid password.';
+
+        $form->addError('password', $errorMessage);
+
+        $this->assertTrue($form->hasErrors());
+        $this->assertEquals($errorMessage, $form->firstError('password'));
+    }
+
     public function testValidatorRules(): void
     {
         $form = new LoginForm();
--- a/tests/FormModelTest.php
+++ b/tests/FormModelTest.php
@@ -141,6 +141,17 @@ final class FormModelTest extends TestCase
         $this->assertEquals($errorMessage, $form->firstError('password'));
     }
 
+    public function testAddAndGetErrorForNonExistingAttribute(): void
+    {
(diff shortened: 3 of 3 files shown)


terabytesoftw:
You can use the validators for that purpose for example to verify username and password:

```php
<?php

public function rules(): array
{
    return [
        'login' => $this->loginRules(),
        'password' => $this->passwordRules($this->userRepository, $this->password)
    ];
}

private function passwordRules(UserRepository $userRepository, ?string $password): array
{
    return [
        new Required(),
        new Callback(
            static function () use ($userRepository, $password): Result {
               return static::validatePassword($password);
            }
       )
    ];
}
```

terabytesoftw:
You can use the validators for that purpose for example to verify username and password:

```php
<?php

public function rules(): array
{
    return [
        'login' => $this->loginRules(),
        'password' => $this->passwordRules($this->userRepository, $this->password)
    ];
}

private function loginRules(): array
{
    return [
        new Required(),
        (new HasLength())
        ->min(4)
        ->max(40)
        ->tooShortMessage('Is too short.')
        ->tooLongMessage('Is too long.'),
        new MatchRegularExpression('/^[-a-zA-Z0-9_\.@]+$/'),
        $this->existValidator($this->userRepository, $this->login)
    ];
}   

private function existValidator(UserRepository $userRepository, ?string $value): Callback
{
    return new Callback(
        static function () use ($userRepository, $value): Result {
            $result = new Result();

            static::$identity = $userRepository->findUserByUsernameOrEmail($value);

            if (static::$identity  === null) {
                $result->addError('Unregistered user.');
            } else {
                if (static::$identity->getAttribute('confirmed_at') === null) {
                    $result->addError('Please check your email to activate your account.');
                }
            }

            return $result;
        }
    );
}

private function passwordRules(UserRepository $userRepository, ?string $password): array
{
    return [
        new Required(),
        new Callback(
            static function () use ($userRepository, $password): Result {
               return static::validatePassword($password);
            }
       )
    ];
}

private static function validatePassword(?string $password): Result
{
    $result = new Result();

    if (static::$identity !== null && !(new PasswordHasher())->validate($password, static::$identity->password_hash)) {
        $result->addError('Invalid password.');
    }

    return $result;
}
```

samdark:
Initially @terabytesoftw added the method and it was me who insisted on removing it. My take is that validation should happen in a single place and form model fits quite well. If there's a need to place DB-specific validation in a separate service, is there a reason to call this service outside of the form model?

armpogart:
@samdark Yes, I have a reason for not placing that service in form model. Here is my use case:
I have a `AuthService` which is responsible for logging in the user, for doing logout, for checking permissions and it can do anything related to authentication or authorization in my case. I don't see a reason to place `AuthService` related logic (validation) in the form and I don't see a reason to duplicate the code or inject `AuthService` both in form and controller, when the form can do pure form validation and only if the form is valid `AuthService` can further check if the user is valid and password is valid in db. And in my opinion it will be more appropriate to even inject `FormModel` to `AuthService` and the service will call the form validation when appropriate, that will allow to call only `AuthService` from controller and the action won't have any unnecessary code in it.
The example you've provided is ok often, but not always. Also one flow I see in it currently is that using `FormModel` for validating everything is not desired usually. I don't want validate all the rules at once, I mean I don't see a need to validate user and the password (through db) if any of that fields failed any previous validation, we just spend precious time unnecessarily by calling db. Do I need to open a separate issue for that?

The other use case I have for this method is the following:
I want to add arbitrary error, such as add error message under some arbitrary attribute name (that doesn't exist), for example `addError('form', 'Username or password is invalid.')` as this error is not for one exact field, but the form overall. The method will allow me to add such arbitrary errors and use it in the view anyway I want.

Insolita:
I consider that this is a good idea! Especially ability to add error for custom attribute names - that sometimes have a sense for API request validations

samdark:
https://forum.yiiframework.com/t/form-should-we-allow-adding-errors-externally/129625

samdark:
https://yiiframework.ru/forum/viewtopic.php?f=39&t=57318

armpogart:
@samdark The last use case is not covered in your polls. How am I supposed to add generic form error and not error directly concerning one exact field. The most obvious example is that on login form when user inputs incorrect username or password it is usually a good practice from security point to not show which of 2 fields were incorrect (not to give sensitive information to any hacker, whether the username exists and password is only incorrect), and we just add generic error message to form that "Username or password is incorrect" in all our projects. And there is no currently anyway not to associate this error to one of the fields.

samdark:
Where in the form would you display that if there's no summary outputted? Usually it is added to either username only or both username and password with the message like "Either username or password is not correct".

FabrizioCaldarelli:
The addError method Is highly useful applied to custom errore attribuite.

In api context I always add an extra attribute named __errors where I put all errors for that object and for the descendendant (to have summary and detail errors).

For example, think at a form to compile an invoice and its rows. In the api create method, I'll have an invoice root node with invoice info and a rowList node with the invoice rows that the user is trying to save.

If there Is some validation error in some row, I'll have __errors property filled with attribute error for that row and a property named 'rowList' (with 'there Is an error in row number xxxx ...' text to display as summary above the invoice form) in __error property of invoice node.

In this way I'll have summary in each container node of the api respose, to easily understand where Is the error in complex forms.

samdark:
OK. Seems majority of votes is for having it.

samdark:
@armpogart would you please make sure it's possible to add/get error for non-existing attribute (general error)?

armpogart:
@samdark Done.

samdark:
Merged. Thank you!

PR #25, opened 2020-06-23
Add the ability to disable hiddeninput in fileinput.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -400,7 +400,7 @@ final class Field extends Widget
      * This method will generate the `name` and `value` tag attributes automatically for the model attribute unless
      * they are explicitly specified in `$options`.
      *
-     * @param bool $noHiddenInput enable/disable hidden input field.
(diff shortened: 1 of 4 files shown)


src/Widget/Field.php
```diff
@@ -410,7 +411,7 @@ public function passwordInput(array $options = []): self
      *
      * @return self the field object itself.
      */
-    public function fileInput(array $options = []): self
+    public function fileInput(bool $noHiddenInput = false, array $options = []): self
```
samdark, line 414:
```suggestion
    public function fileInput(bool $withoutHiddenInput = false, array $options = []): self
```

Also it may make sense to swap arguments order:

```suggestion
    public function fileInput(array $options = [], bool $withoutHiddenInput = false): self
```

terabytesoftw: Done.

PR #26, opened 2020-06-25
Improve detecting form name

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -77,9 +77,14 @@ abstract class FormModel implements FormModelInterface
         return $this->attributesLabels[$attribute] ?? $this->generateAttributeLabel($attribute);
     }
 
+    /**
+     * @return string Returns classname without a namespace part or empty string when class is anonymous
+     */
     public function formName(): string
     {
-        return '';
+        return strpos(static::class, '@anonymous') !== false
+            ? ''
+            : mb_substr(mb_strrchr(static::class, "\\"), 1);
     }
 
     public function hasAttribute(string $attribute): bool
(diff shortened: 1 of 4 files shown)


src/FormModel.php
```diff
@@ -77,9 +77,18 @@ public function attributeLabel(string $attribute): string
...
     public function formName(): string
     {
-        return '';
+        $className = strrchr(static::class, "\\");
+        if ($className === false) {
+            // it's anonymous class
+            return '';
+        }
+
+        return substr($className ?: '', 1);
```
roxblnfk, line 91:
```suggestion
        return $className === false ? '' : substr($className, 1);
```

terabytesoftw:
Test fails PHP 8.0

xepozz:
Anybody know why?

src/FormModel.php
```diff
@@ -260,6 +265,7 @@ private function addErrors(array $items): void
     private function attributes(): array
     {
         $class = new \ReflectionClass($this);
+        $this->attributes = [];
```
samdark, line 268: Looks like side effect.is it necessary to modify the form in a query method?

xepozz: [This line](https://github.com/yiisoft/form/blob/edfeaf089d54cb9865ad4d7e634dccb170b63d8d/src/FormModel.php#L284) read the property, but if we have no one property it will throw error.

samdark:
:+1:

PR #27, opened 2020-06-27
Small improvements

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -33,7 +33,7 @@ abstract class FormModel implements FormModelInterface
         $this->translationDomain = $translationDomain;
         $this->translationLocale = $translationLocale;
 
-        $this->attributes = $this->attributes();
+        $this->attributes = $this->collectAttributes();
         $this->attributesLabels = $this->attributeLabels();
     }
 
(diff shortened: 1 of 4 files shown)


src/FormModel.php
```diff
@@ -262,26 +262,28 @@ private function addErrors(array $items): void
...
-
-                if ($type !== null) {
-                    $this->attributes[$property->getName()] = $type->getName();
-                } else {
-                    throw new \InvalidArgumentException(
-                        "You must specify the type hint for \"$property->class\" class."
-                    );
-                }
+            if ($property->isStatic()) {
+                continue;
```
terabytesoftw, line 272: You are adding a mutant, just check that it is not static.

xepozz: What?

samdark: Likely it's about mutation testing and Infection.

terabytesoftw: https://infection.github.io/guide/mutators.html#Loop

xepozz: Foreach and for produces mutations too. We won't use it too?

xepozz: image
I've fixed it.

xepozz:
I've fixed some other mutants too

samdark:
Thanks!

Issue #17, opened 2020-05-19
Form fields are not customizable


armpogart:
I like in which direction this library goes, but I still don't like some aspects of it, will post each problem in a separate issue. This one concerning `\Yiisoft\Yii\Form\Widget\Field` class extendibility.

In yii2 all widgets were just not final classes and our own classes could easily extend them to customize the behavior wanted. I'm aware of the issues that not final classes pose, such as it becomes more difficult to maintain/support/change them later as each change can cause BC break, so here comes `final` classes. And I always encourage to use `final` classes where possible. But Meyer's open-closed principle (from SOLID) still states that the class must be open for extension (while still being closed for modification). DI in yii3 partly solves the problem, I say partly as the classes must be still designed carefully to allow that extendibility. And now comes an example.

I want to customize the fields for usage with bootstrap 4, here is an example of a typical bootstrap 4 form:
```html
<form>
  <div class="form-group">
    <label for="inputLogin">Login</label>
    <div class="input-group">
      <div class="input-group-prepend">
        <span class="input-group-text" id="inputGroupPrepend">@</span>
      </div>
      <input type="text" class="form-control" id="inputLogin" placeholder="Example input placeholder">
    </div>
  </div>
  <div class="form-group">
    <label for="inputPassword">Another label</label>
    <input type="text" class="form-control" id="inputPassword" placeholder="Another input placeholder">
    <small class="form-text text-muted">
      Your password must be 8-20 characters long, contain letters and numbers, and must not contain spaces, special characters, or emoji.
    </small>
  </div>
  <div class="form-group">
    <label for="inputSelect">Select Label</label>
    <select id="inputSelect" class="form-control">
        <option selected>Choose...</option>
        <option>...</option>
    </select>
  </div>
</form>
```

Now the issues with current implementation:
  - I want to remove default label class (`\Yiisoft\Yii\Form\Widget\Field::DEFAULT_LABEL_OPTIONS`) from all fields, but I can't do that currently as even though `Field` is a widget and is configured through factory, but I can't set private constant through factory.
  - I want to use `form-text` or `text-muted` or both classes for instead of default `hint-block`, and also use `small` tag instead of `div` by default, but again I don't see any way to configure private property ((`\Yiisoft\Yii\Form\Widget\Field::DEFAULT_HINT_OPTIONS`) or `options` through factory.
  - I want to apply `custom-select` class to all dropdown lists (it is important for bootstrap 4 select validation styling) and again I don't see how.
[...]

PR #33, opened 2020-08-12
Add an ability to disable container for the field.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | Fixes one of the points (7) mentioned in [#17](https://github.com/yiisoft/form/issues/17#issuecomment-672885901)

### DIFF
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -43,6 +43,7 @@ final class Field extends Widget
     private ?string $inputId = null;
     private array $parts = [];
     private bool $skipForInLabel = false;
+    private bool $containerEnabled;
 
     /**
      * Renders the whole field.
@@ -83,7 +84,12 @@ final class Field extends Widget
             $content = $content($this);
         }
 
-        return $this->renderBegin() . "\n" . $content . "\n" . $this->renderEnd();
+        if ($this->containerEnabled) {
+            return $this->renderBegin() . "\n" . $content . "\n" . $this->renderEnd();
+        }
+
+        return $content;
+
     }
 
     /**
@@ -741,17 +747,19 @@ final class Field extends Widget
      *
      * @param FormModelInterface $data Form model.
      * @param string $attribute Form model property this widget is rendered for.
-     * @param array $options The HTML attributes for the widget container tag.
+     * @param array $options The HTML attributes for the widget container tag if it is enabled.
      * See {@see \Yiisoft\Html\Html::renderTagAttributes()} for details on how attributes are being rendered.
+     * @param bool $containerEnabled whether the widget container is enabled. Default: true
      *
      * @return self
      */
-    public function config(FormModelInterface $data, string $attribute, array $options = []): self
+    public function config(FormModelInterface $data, string $attribute, array $options = [], $containerEnabled = true): self
     {
         $new = clone $this;
         $new->data = $data;
         $new->attribute = $attribute;
         $new->options = $options;
+        $new->containerEnabled = $containerEnabled;
         return $new;
     }
 
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -743,21 +743,32 @@ final class Field extends Widget
     }
 
     /**
-     * Set form model, name and options for the widget.
+     * Set form model and name for the widget.
      *
      * @param FormModelInterface $data Form model.
      * @param string $attribute Form model property this widget is rendered for.
-     * @param array $options The HTML attributes for the widget container tag if it is enabled.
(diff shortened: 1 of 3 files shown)


src/Widget/Field.php
```diff
@@ -741,17 +747,19 @@ private function addInputId(): string
...
      * @param string $attribute Form model property this widget is rendered for.
-     * @param array $options The HTML attributes for the widget container tag.
+     * @param array $options The HTML attributes for the widget container tag if it is enabled.
      * See {@see \Yiisoft\Html\Html::renderTagAttributes()} for details on how attributes are being rendered.
+     * @param bool $containerEnabled whether the widget container is enabled. Default: true
      *
      * @return self
      */
-    public function config(FormModelInterface $data, string $attribute, array $options = []): self
+    public function config(FormModelInterface $data, string $attribute, array $options = [], $containerEnabled = true): self
```
terabytesoftw, line 756: I think a containerEnabled() method would be better.

armpogart: @terabytesoftw I first though also about adding a method, but it would be weird to have an ability to do `containerEnabled(true)` if it is already enabled by default. It would make sense to have a separate method if we also `$options` there too. In that case we will remove `$options` from `config` method and add `container` method with 2 arguments `function container($enabled = true, array $options = [])`. What do you think?

armpogart: If `$options` for the container are configured in `config()` method and enabling/disabling it is a separate method, it becomes weird to have an ability to set the options in config and then disable the container. So I think it is better to keep these 2 things together either in `config` method as it is now or a separate method with both of them.

terabytesoftw: We can change the name to `enclosedByContainer()` and in the same way pass the options for the container.

src/Widget/Field.php
```diff
@@ -737,21 +744,34 @@ private function addInputId(): string
...
+     *
+     * @param bool $containerEnabled enabled/disable container for the widget.
+     * @param array $options The HTML attributes for the widget container tag if it is enabled.
+     * See {@see \Yiisoft\Html\Html::renderTagAttributes()} for details on how attributes are being rendered.
+     * @return self
+     */
+    public function enclosedByContainer($containerEnabled = true, array $options = []): self
+    {
+        $new = clone $this;
         $new->options = $options;
```
xepozz, line 773: Why did you move `$options` here?

src/Widget/Field.php
```diff
@@ -737,21 +744,34 @@ private function addInputId(): string
...
+
+    /**
+     * Generate a container tag for {@see attribute}.
+     *
+     * @param bool $containerEnabled enabled/disable container for the widget.
+     * @param array $options The HTML attributes for the widget container tag if it is enabled.
+     * See {@see \Yiisoft\Html\Html::renderTagAttributes()} for details on how attributes are being rendered.
+     * @return self
+     */
+    public function enclosedByContainer($containerEnabled = true, array $options = []): self
```
xepozz, line 770:
```suggestion
    public function enclosedByContainer(bool $containerEnabled, array $options = []): self
```

src/Widget/Field.php
armpogart: See [discussion](https://github.com/yiisoft/form/pull/33#discussion_r469294033) on @terabytesoftw requested changes above.

xepozz: What about to split these onto 2 independent methods?

armpogart: Do you mean something like `containerEnabled(bool $enabled)` and `containerOptions(array $options = [])`?
I have tried to replicate `label(bool $enabledLabel = true, array $options = [], ?string $label = null)` method, which is already there and combines enabling/disabling label with its' options.

armpogart: I'm ok with any approach allowing to disable container, but I would like to keep behaviors persistent. So in case of 2 independent methods, the same thing must be done for label.

xepozz: Ok let it be. We can think about it later.

samdark: It can be split to withoutContainer() and withContainer(array $options).

armpogart: @samdark What about label in that case?

terabytesoftw:
👍

src/Widget/Field.php
samdark: That's just a method split nothing is changed about label.

armpogart: @samdark I mean the same must be done with `label` method (which works the same now) to be consistent.

PR #34, opened 2020-08-14
Directly read/write public properties without binding closure.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Don't use closure binding when reading/writing public properties (attributes).


terabytesoftw:
👍

PR #38, opened 2020-08-14
Use getter/setter for attributes if available.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Use getter/setter for reading and writing attributes if they are available. Use case: when you pre-processing/formatting beforehand.
Also replaced `get_class($this)` with `static::class` as it is faster with less op calls: https://belineperspectives.com/2017/03/13/get_classthis-vs-staticclass/

P.S. Will add/fix tests after initial review.

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -338,7 +338,7 @@ abstract class FormModel implements FormModelInterface
 
         $getter = 'get' . $this->inflector->camelize($attribute);
         if (is_callable([$this, $getter], true)) {
-            $method = new ReflectionMethod($class, 'get' . $this->inflector->camelize($attribute));
+            $method = new ReflectionMethod($class, $getter);
             if ($method->getNumberOfRequiredParameters() === 0) {
                 return $this->$getter();
             }
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -336,8 +336,8 @@ abstract class FormModel implements FormModelInterface
             throw new InvalidArgumentException("Undefined property: \"$class::$attribute\".");
         }
 
-        $getter = 'get' . $this->inflector->camelize($attribute);
-        if (is_callable([$this, $getter], true)) {
+        $getter = 'get' . $this->getInflector()->camelize($attribute) . 'Attribute';
+        if (is_callable([$this, $getter])) {
             $method = new ReflectionMethod($class, $getter);
             if ($method->getNumberOfRequiredParameters() === 0) {
                 return $this->$getter();
@@ -356,13 +356,16 @@ abstract class FormModel implements FormModelInterface
 
     private function writeProperty(string $attribute, $value): void
     {
-        $setter = 'set' . $this->inflector->camelize($attribute);
+        $setter = 'set' . $this->getInflector()->camelize($attribute) . 'Attribute';
         if (is_callable([$this, $setter])) {
             $method = new ReflectionMethod(static::class, $setter);
             if ($method->getNumberOfRequiredParameters() === 1) {
                 $this->$setter($value);
+                return;
             }
-        } else if ($this->isPublicAttribute($attribute)) {
+        }
+
+        if ($this->isPublicAttribute($attribute)) {
             $this->$attribute = $value;
         } else {
             $setter = fn($class, $attribute, $value) => $class->$attribute = $value;
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -7,6 +7,7 @@ namespace Yiisoft\Form;
 use Closure;
 use InvalidArgumentException;
 use ReflectionClass;
+use ReflectionMethod;
 use Yiisoft\Strings\Inflector;
 use Yiisoft\Validator\Rule\Required;
 use Yiisoft\Validator\ValidatorFactoryInterface;
@@ -14,6 +15,8 @@ use Yiisoft\Validator\ValidatorFactoryInterface;
 use function array_key_exists;
 use function array_merge;
 use function get_object_vars;
+use function is_callable;
+use function property_exists;
 use function reset;
 use function sprintf;
 
@@ -328,9 +331,17 @@ abstract class FormModel implements FormModelInterface
 
     private function readProperty(string $attribute)
     {
-        $class = get_class($this);
+        $class = static::class;
         if (!property_exists($class, $attribute)) {
             throw new InvalidArgumentException("Undefined property: \"$class::$attribute\".");
+        }
+
+        $getter = 'get' . $this->inflector->camelize($attribute);
+        if (is_callable([$this, $getter], true)) {
+            $method = new ReflectionMethod($class, 'get' . $this->inflector->camelize($attribute));
+            if ($method->getNumberOfRequiredParameters() === 0) {
+                return $this->$getter();
+            }
         }
 
         if ($this->isPublicAttribute($attribute)) {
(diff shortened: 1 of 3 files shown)


src/FormModel.php
```diff
@@ -328,11 +331,19 @@ private function generateAttributeLabel(string $name): string
 
     private function readProperty(string $attribute)
     {
-        $class = get_class($this);
+        $class = static::class;
         if (!property_exists($class, $attribute)) {
```
xepozz, line 335: Shouldn't you move it below?

src/FormModel.php
```diff
@@ -328,11 +331,19 @@ private function generateAttributeLabel(string $name): string
...
             throw new InvalidArgumentException("Undefined property: \"$class::$attribute\".");
         }
 
+        $getter = 'get' . $this->inflector->camelize($attribute);
+        if (is_callable([$this, $getter], true)) {
+            $method = new ReflectionMethod($class, 'get' . $this->inflector->camelize($attribute));
```
xepozz, line 341:
```suggestion
            $method = new ReflectionMethod($class, $getter);
```

src/FormModel.php
```diff
@@ -345,7 +356,13 @@ private function readProperty(string $attribute)
...
     private function writeProperty(string $attribute, $value): void
     {
-        if ($this->isPublicAttribute($attribute)) {
+        $setter = 'set' . $this->inflector->camelize($attribute);
+        if (is_callable([$this, $setter])) {
+            $method = new ReflectionMethod(static::class, $setter);
+            if ($method->getNumberOfRequiredParameters() === 1) {
+                $this->$setter($value);
+            }
+        } else if ($this->isPublicAttribute($attribute)) {
```
xepozz, line 365:
```suggestion
        }
        if ($this->isPublicAttribute($attribute)) {
```

src/FormModel.php
armpogart: Yep, missed that one.

src/FormModel.php
armpogart: The consequence of drafting and not refactoring correctly using IDE tools :)

src/FormModel.php
armpogart: Ideally we need `else if`, but I see the problem. The first `if` can return true and then execution will be exited when checking for parameters. So I need also to return if nested `if` did the work.

xepozz: Why the second `return` is needed when count parameters is not equals to 1?
Maybe just avoid to checking it?

src/FormModel.php
armpogart: While looking again on it.... Do we really want to allow getter and setter to work without having attribute at all? `writeProperty` checks for property existence anyway as it is called directly only in `setAttributes` where this check is done.

armpogart: I can surely refactor all that to ignore attribute (property) existence in both cases, and `collectAttributes` can also collect getters and setters in that case. But in that case we won't know whether the intent of getter/setter is to load/get property or some other use case just for model logic.

src/FormModel.php
rustamwin, line 365:
```suggestion
        } elseif ($this->isPublicAttribute($attribute)) {
```

src/FormModel.php
xepozz: I think we need to use `set{field name}` even if property doesn't exists.
For example property can be named with underscore that reflect private property (old school naming)

armpogart: Yep, but what if I add some setter or getter for internal logic only, see for example `getInflector` method. In this exact case it is private and will be ignored, but I also have public such public method which is only used for mapping to entity for example. In that case we will need another static method to define which attributes (properties) are used in these process and which are not.

xepozz: internal methods should be private

src/FormModel.php
armpogart: If I replace `else if` with `if` than if setter is available and correct (has required parameter) it will be called, after which also property will be written directly as we don't terminate execution. On the other if don't check for parameters we can have `setParameter()` method which has the correct naming, but is not intended to be used for setting or some other thing internally in the application, and consequently it doesn't have 1 required parameter to be used as setter and will have runtime exception. With the check we will just ignore it and fallback to setting attribute directly.

armpogart: The resolution will depend on the result of [discussion](https://github.com/yiisoft/form/pull/38#discussion_r470771047) above.

armpogart: I mean something like:
```php
if (is_callable([$this, $setter])) {
    $method = new ReflectionMethod(static::class, $setter);
    if ($method->getNumberOfRequiredParameters() === 1) {
        $this->$setter($value);
        return;
    }
}

if ($this->isPublicAttribute($attribute)) {
    $this->$attribute = $value;
} else {
    $setter = fn($class, $attribute, $value) => $class->$attribute = $value;
    $setter = Closure::bind($setter, null, $this);

    $setter($this, $attribute, $value);
}
```

src/FormModel.php
armpogart: I think it will be better to open a separate issue to discuss the usability of allowing getter/setter without actual attribute considering the potential problem I have mentioned above. Also it would be better to address that as a separate PR as it would need a rewrite not only to `readProperty`/`writeProperty` methods, but also `collectAttributes` to collect virtual (getter and setter only without actual attribute) attributes by reading method signature (parameter type of setter or return type of getter). I see that can be helpful sometimes not to have backing attribute at all and have virtual attribute which is just composed attribute for example, e.g. `$this->firstName . ' ' . $this->lastName`. But in that case why do we need `firstName` and `lastName` properties on model at all? I will address all issues that come to my mind in that separate issue.

armpogart: > internal methods should be private

By internal I mean internal to application logic not the class itself. Some `getParam` which is used only in a service that is doing mapping form to entity. The method will be public in that case, but it won't have any meaning in the context of setting/getting parameters.

armpogart: By allowing getter/setter without actual property we are just restricting the naming of methods. We are disallowing the end user to name any method `getSomething` or `setSomething` if the method is not intended to be used in attribute collection process, so we need to weigh out on that.

armpogart:
The first iteration was using `getPropertyName` and `setPropertyName` directly, now it uses `getPropertyNameAttribute` and `setPropertyNameAttribute` alternative getter and setter to allow to have normal getter and setter in the model that is not participating in reading/writing properties from `FormModel::load` and Html form field which uses `getAttributeValue` context.

There is also discussion and possibility to extend this to allow even getter and setter without actual property, some sort of virtual attributes for that context.

But I am not sure whether this functionality is needed in model at all. Something similar was present in Yii2, but it was usable there as `ActiveRecord` model was often used as an actual model for forms and loading from request. Here `FormModel` and actually DTO and it's not correct to do formatting/pre-processing in DTO. It is better done on the stage of mapping entity to dto and vice-versa.

Leaving my thoughts and PR here for further discussion.

P.S. In case if PR closed at the end without merging, I will push the change of `get_class` to `static::class` with another PR.

armpogart:
In any case we can document the possibility of doing something similar with just overriding `setAttribute` (currently there is only `setAttributes`, but I will make it private and add public `setAttribute` instead)  and `getAttributeValue` in extending class.

src/FormModel.php
xepozz: > I think it will be better to open a separate issue

Yes, you are right. #39

src/FormModel.php
xepozz: When `elseif` will remain the property will intact if form will have not correct setter.

I think this mechanism should work by this rules:
1. Use setter if available
2. Use directly property

samdark:
@armpogart would you please resolve conflicts? Is there anything to add/fix in this pull request?

samdark:
See issue for reason.

PR #40, opened 2020-08-15
Removing invalid enctype and value from file input

Mister-42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  |


samdark:
Thanks!

Mister-42:
What is the function of the hidden input? 🤔

Issue #7, opened 2020-04-27
Add documentation.


terabytesoftw:
Add documentation:

- [ ] all widgets.
- [ ] forms.

samdark:
- [ ] How to use translation.

vjik:
Need example php code to generate for cases like this:

<input " type="text" name="LocationForm[coords][lng]">

PR #43, opened 2020-08-17
Support for nested attributes.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | reference [#17](https://github.com/yiisoft/form/issues/17#issuecomment-671545891)

The only reason this PR also breaks BC is that I have removed public `setAttributes` method from `FormModelInterface` and `FormModel` as it didn't have any reason to be there. The same thing could have been done with `load`. Even documentation for `load` was saying the same thing, but wasn't supporting that behavior. I have also added new public method to the interface and class `setAttribute(string $name, $value)` which encapsulated some of the logic from `setAttributes` that was removed and this new method is used in my implementation of nested attributes.

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -13,9 +13,12 @@ use Yiisoft\Validator\ValidatorFactoryInterface;
 
 use function array_key_exists;
 use function array_merge;
+use function explode;
 use function get_object_vars;
+use function is_subclass_of;
 use function reset;
 use function sprintf;
+use function strpos;
 
 /**
  * Form model represents an HTML form: its data, validation and presentation.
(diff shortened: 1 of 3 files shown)


armpogart:
Also docs are missing, there is already issue #7 for adding overall documentation.

armpogart:
The current format for nested attribute is `attribute.nestedAttribute`. I was considering `attribute[nestedAttribute]` as that would allow easier usage in API context (JSON), but I haven't looked through whether that format will conflict with array attributes (if they are supported at all). So maybe in the future :)

armpogart:
I don't like how nested attribute labels and hints are got. But would prefer to refactor that as a separate PR with issue. There is some things that need discussion beforehand.

samdark (approved):

samdark:
👍

armpogart:
@samdark Oops, you merged it :) the validation and errors are not working for nested attributes, I wrote to @terabytesoftw earlier about that and forgot to mention it here. I was finishing the work on it, to add another commit. Will open a new PR with added parts and some refactoring for the merged code.

PR #49, opened 2020-10-21
Adopt Psalm

Fantom409:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ❌
| Fixed issues  |


src/FormModel.php
```diff
@@ -150,10 +150,18 @@ public function hasErrors(?string $attribute = null): bool
         return $attribute === null ? !empty($this->attributesErrors) : isset($this->attributesErrors[$attribute]);
     }
 
+    /**
+     * @param array $data
+     * @param string|null $formName
+     * @return bool
+     */
     public function load(array $data, $formName = null): bool
```
samdark, line 158: Can't we adjust signature itself and remove phpdoc here?

```suggestion
    public function load(array $data, ?string $formName = null): bool
```

src/Helper/HtmlForm.php
```diff
@@ -51,7 +51,7 @@ public static function getAttributeValue(FormModelInterface $form, string $attri
      */
     public static function getInputId(FormModelInterface $form, string $attribute, string $charset = 'UTF-8'): string
     {
-        $name = mb_strtolower(static::getInputName($form, $attribute), $charset);
+        $name = (string) mb_strtolower(static::getInputName($form, $attribute), $charset);
```
samdark, line 54: I don't think it can return non-string.

src/Widget/Form.php
```diff
@@ -50,7 +51,7 @@ public function start(): string
                 if (($pos1 = strpos($pair, '=')) !== false) {
                     $hiddenInputs[] = Html::hiddenInput(
                         urldecode(substr($pair, 0, $pos1)),
-                        urldecode(substr($pair, $pos1 + 1))
+                        urldecode((string) substr($pair, $pos1 + 1))
```
samdark, line 54: While it theoretically can return `false`, I see it as excessive type-casting in this and alike cases.

samdark (changes requested):

src/Helper/HtmlForm.php
Fantom409: > I don't think it can return non-string.

mb_strtolower return false and generate a warning if is called with undefined charset (i.e. mb_strtolower('example', 'unkownCharset') )

samdark: Why should we cast this `false` to a string?

samdark (approved):

PR #61, opened 2020-12-14
Add required class to labels

Mister-42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -881,6 +881,10 @@ final class Field extends Widget
 
     public function labelOptions(array $options = []): void
     {
+        if ($this->data->isAttributeRequired($this->attribute)) {
+            Html::addCssClass($options, $this->requiredCssClass);
+        }
+
         $classFromOptions = ArrayHelper::remove($options, 'class', self::DEFAULT_LABEL_OPTIONS['class']);
 
         if ($classFromOptions !== self::DEFAULT_LABEL_OPTIONS['class']) {
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -157,6 +157,10 @@ final class Field extends Widget
             $new->labelOptions['label'] = $label;
         }
 
+        if ($new->data->isAttributeRequired($new->attribute)) {
+            Html::addCssClass($options, $new->requiredCssClass);
+        }
+
         $new->labelOptions($options);
         $new->skipForInLabel();
 
@@ -881,10 +885,6 @@ final class Field extends Widget
 
     public function labelOptions(array $options = []): void
     {
-        if ($this->data->isAttributeRequired($this->attribute)) {
-            Html::addCssClass($options, $this->requiredCssClass);
-        }
-
         $classFromOptions = ArrayHelper::remove($options, 'class', self::DEFAULT_LABEL_OPTIONS['class']);
 
         if ($classFromOptions !== self::DEFAULT_LABEL_OPTIONS['class']) {
--- a/tests/Widget/FieldErrorTest.php
+++ b/tests/Widget/FieldErrorTest.php
@@ -18,7 +18,7 @@ final class FieldErrorTest extends TestCase
 
         $expected = <<<'HTML'
 <div class="form-group field-personalform-name">
-<label class="control-label" for="personalform-name">Name</label>
+<label class="control-label required" for="personalform-name">Name</label>
 <input type="text" id="personalform-name" class="form-control has-error" name="PersonalForm[name]" value="yii" aria-required="true" aria-invalid="true" placeholder="Name">
 <div class="hint-block">Write your first name.</div>
 <div class="help-block">Is too short.</div>
@@ -39,7 +39,7 @@ HTML;
 
         $expected = <<<'HTML'
 <div class="form-group field-personalform-name">
-<label class="control-label" for="personalform-name">Name</label>
+<label class="control-label required" for="personalform-name">Name</label>
 <input type="text" id="personalform-name" class="form-control has-error" name="PersonalForm[name]" value="yii" aria-required="true" aria-invalid="true" placeholder="Name">
 <div class="hint-block">Write your first name.</div>
 <div class="help-block errorTestMe">Is too short.</div>
--- a/tests/Widget/FieldHintTest.php
+++ b/tests/Widget/FieldHintTest.php
@@ -16,7 +16,7 @@ final class FieldHintTest extends TestCase
 
         $expected = <<<'HTML'
 <div class="form-group field-personalform-name">
-<label class="control-label" for="personalform-name">Name</label>
+<label class="control-label required" for="personalform-name">Name</label>
 <input type="text" id="personalform-name" class="form-control" name="PersonalForm[name]" aria-required="true" placeholder="Name">
 <div class="hint-block">Write your first name.</div>
 <div class="help-block"></div>
@@ -34,7 +34,7 @@ HTML;
 
         $expected = <<<'HTML'
 <div class="form-group field-personalform-name">
-<label class="control-label" for="personalform-name">Name</label>
+<label class="control-label required" for="personalform-name">Name</label>
 <input type="text" id="personalform-name" class="form-control" name="PersonalForm[name]" aria-required="true" placeholder="Name">
 <div class="hint-block customClass">Custom hint.</div>
 <div class="help-block"></div>
@@ -53,7 +53,7 @@ HTML;
 
         $expected = <<<'HTML'
 <div class="form-group field-personalform-name">
-<label class="control-label" for="personalform-name">Name</label>
+<label class="control-label required" for="personalform-name">Name</label>
 <input type="text" id="personalform-name" class="form-control" name="PersonalForm[name]" aria-required="true" placeholder="Name">
 
 <div class="help-block"></div>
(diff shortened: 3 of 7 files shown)


Mister-42:
Still got an error in real web environment after tests ran perfectly.

terabytesoftw:
Could you show how to reproduce the error ?

Thks,

Mister-42:
```php
echo Form::widget()
    ->begin();
```

That triggered the error `Typed property Yiisoft\Form\Widget\Field::$data must not be accessed before initialization`.

terabytesoftw:
It's weird because `Form::class` is not related to `Field::class`, add tests form empty and works, https://github.com/yiisoft/form/pull/62, it must be another code, from `Field::class` in the view.

Mister-42:
Controller
```php
    public function weeknumbers(): ResponseInterface
    {
        return $this->viewRenderer->renderPartial('weeknumbers'); //changed to partial for test to rule out layout
    }
```

view
```php
<?php

declare(strict_types=1);

use Yiisoft\Form\Widget\Form;

echo Form::widget()->begin;
exit;
```

Results:
`Typed property Yiisoft\Form\Widget\Field::$data must not be accessed before initialization`

Mister-42:
The issue does not occur after second commit, but I found it very strange that tests does not show this behaviour.

terabytesoftw:
> Controlador
> 
> ```
>     public function weeknumbers(): ResponseInterface
>     {
>         return $this->viewRenderer->renderPartial('weeknumbers'); //changed to partial for test to rule out layout
>     }
> ```
> 
> Vista
> 
> ```
> <?php
> 
> declare(strict_types=1);
> 
> use Yiisoft\Form\Widget\Form;
> 
> echo Form::widget()->begin;
> exit;
> ```
> 
> Resultados:
> `Typed property Yiisoft\Form\Widget\Field::$data must not be accessed before initialization`

you are using `begin` without parentheses, its `echo Form::widget()->begin();`

Mister-42:
Same result

terabytesoftw:
t's weird because I literally do the same thing in the tests.

Mister-42:
Yeah I noticed. Except there is a viewRenderer in between on the web.

Mister-42:
- install yii-demo and run `composer u`
- replace vendor/yiisoft/form/src/Widget/Field.php with [this version](https://github.com/yiisoft/form/raw/41492af49b4a750f4963f0d3ee5c2f397fb3165c/src/Widget/Field.php)
- `vendor/bin/yii serve`
- visit http://localhost:8080/contact

There is the error

Mister-42:
> It's weird because `Form::class` is not related to `Field::class`, add tests form empty and works, #62, it must be another code, from `Field::class` in the view.

Field is loaded by [config](https://github.com/yiisoft/yii-demo/blob/master/config/web/fields.php).

Mister-42:
@samdark Current code works in web and in tests and can be reviewed.

terabytesoftw:
It is not necessary to changeof the widgets to add a class, it is precisely like that, so that it supports any css framework.

```php
$fieldConfig = [
    'labelOptions()' => [['class' => 'required'],
];

Form::widget() ->begin();
    Field::Widget($fieldConfig)->config($form, 'username')
Form::end();
```

terabytesoftw:
> > It's weird because `Form::class` is not related to `Field::class`, add tests form empty and works, #62, it must be another code, from `Field::class` in the view.
> 
> Field is loaded by [config](https://github.com/yiisoft/yii-demo/blob/master/config/web/fields.php).

You have an error in the configuration, you can place it.

Mister-42:
> It is not necessary to change of the widgets to add a class, it is precisely like that, so that it supports any css framework.
> 
> ```
> $fieldConfig = [
>     'labelOptions()' => [['class' => 'required'],
> ];
> 
> Form::widget()->begin();
>     Field::Widget($fieldConfig)->config($form, 'username')
> Form::end();
> ```
> 
> I am against adding classes by default, only the minimum ones we have.

You want ALL fields marked as required? That is not what I am doing here...
The `required` class is very generic and can be used in any css framework. It's not even defined bootstrap css.

Mister-42:
> > > It's weird because `Form::class` is not related to `Field::class`, add tests form empty and works, #62, it must be another code, from `Field::class` in the view.
> > 
> > 
> > Field is loaded by [config](https://github.com/yiisoft/yii-demo/blob/master/config/web/fields.php).
> 
> You have an error in the configuration, you can place it.

That's not my config, that's yii-demo

terabytesoftw:
config global in config/fields:

```php
<?php

declare(strict_types=1);

use Yiisoft\Form\Widget\Field;

/** @var array $params */

return [
    Field::class => static function () {
        $fieldConfig = [
            'inputOptions()' => [['required' => true]],
        ];

        return Field::Widget($fieldConfig);
    }
];
```

Mister-42:
> config global in config/fields:
> 
> ```
> <?php
> 
> declare(strict_types=1);
> 
> use Yiisoft\Form\Widget\Field;
> 
> /** @var array $params */
> 
> return [
>     Field::class => static function () {
>         $fieldConfig = [
>             'inputOptions()' => [['required' => true]],
>         ];
> 
>         return Field::Widget($fieldConfig);
>     }
> ];
> ```

That is a terrible idea. This PR adds `required` only when it is marked as required field by `Yiisoft\Validator\Rule\Required`

terabytesoftw:
> > config global in config/fields:
> > ```
> > <?php
> > 
> > declare(strict_types=1);
> > 
> > use Yiisoft\Form\Widget\Field;
> > 
> > /** @var array $params */
> > 
> > return [
> >     Field::class => static function () {
> >         $fieldConfig = [
> >             'inputOptions()' => [['required' => true]],
> >         ];
> > 
> >         return Field::Widget($fieldConfig);
> >     }
> > ];
> > ```
> 
> That is a terrible idea. This PR adds `required` only when it is marked as required field by `Yiisoft\Validator\Rule\Required`

What sense does it make to place the validator and the required option, they are two very different validations, anyway as you have done the pr it does not work, try it.

Mister-42:
> What sense does it make to place the validator and the required option, they are two very different validations

If it marked by validator as required, it is a required field. I can't make more sense about that than this.

> anyway as you have done the pr it does not work, try it.

Does work, see tests

terabytesoftw:
> > What sense does it make to place the validator and the required option, they are two very different validations
> 
> If it marked by validator as required, it is a required field. I can't make more sense about that than this.
> 
> > anyway as you have done the pr it does not work, try it.
> 
> Does work, see tests

You are putting a required class in label that does nothing, must go the attribute `required` in <input required>

Mister-42:
> Example:
> 
> ```
> <input type="text" id="contactform-username" class="form-control" name="ContactForm[username]" value="" required="" aria-required="true" placeholder="Username">
> ```

No, you got this PR all wrong

```php
<label class="control-label required" for="contactform-username">Name</label>
 <input type="text" id="contactform-username" class="form-control" name="ContactForm[username]" value="" aria-required="true" placeholder="Username">
```

terabytesoftw:
> > Example:
> > ```
> > <input type="text" id="contactform-username" class="form-control" name="ContactForm[username]" value="" required="" aria-required="true" placeholder="Username">
> > ```
> 
> No, you got this PR all wrong
> 
> ```
> <label class="control-label required" for="contactform-username">Name</label>
>  <input type="text" id="contactform-username" class="form-control" name="ContactForm[username]" value="" aria-required="true" placeholder="Username">
> ```
> 
> > You are putting a required class in label that does nothing, must go the attribute `required` in `<input required>` , and for me it does not make sense, because the browser will validate before the yii validator.
> 
> So the label **can** by styled, giving a visual clue of that fields are required.
> 
> Example:
> 
> ```css
> .required:after {
>     color: $red;
>     content: '\a0*';
> }
> ```

I already understood I thought you wanted to validate with the browser, but you can still add your class in the general config:

in Params:

```php
    'yiisoft/form' => [
        'fieldConfig' => [
            'inputCssClass()' => ['form-control input field'],
            'labelOptions()' => [['class' => 'required', 'label' => '']],
        ],
    ],
```
config global in config/fields:

```php
<?php

declare(strict_types=1);

use Yiisoft\Form\Widget\Field;

/** @var array $params */

return [
    Field::class => static fn () => Field::Widget($params['yiisoft/form']['fieldConfig']),
];
```

Mister-42:
> but you can still add your class in the general config:

No, because that will mark __all__ labels as required, not just the labels belonging to required fields.

terabytesoftw:
> > but you can still add your class in the general config:
> 
> No, because that will mark **all** labels as required, not just the labels belonging to required fields.

So I already understood and if it now makes sense to me.

src/Widget/Field.php
```diff
@@ -157,6 +157,10 @@ public function label(bool $enabledLabel = true, array $options = [], ?string $l
             $new->labelOptions['label'] = $label;
         }
 
+        if ($new->data->isAttributeRequired($new->attribute)) {
+            Html::addCssClass($options, $new->requiredCssClass);
```
samdark, line 161: Maybe it's a better idea to add HTML5 validation attribute here? What do you think?

See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

Mister-42: That would be an attribute on the input, so browser does validation. I do class on label for visual indication for end user of required field. Different use case.

Mister-42: ` aria-required="true"` in the input is already btw

samdark:
👍

PR #64, opened 2020-12-22
Fix add success class to field.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

- Just add the has-success class, if there are no errors and the validator was run.

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -32,7 +32,7 @@ abstract class FormModel implements FormModelInterface
     private array $attributesLabels;
     private array $attributesErrors = [];
     private ?Inflector $inflector = null;
-    private bool $isValidate = false;
+    private bool $validated = false;
 
     public function __construct(ValidatorFactoryInterface $validatorFactory)
     {
@@ -232,7 +232,7 @@ abstract class FormModel implements FormModelInterface
             }
         }
 
-        $this->isValidate = true;
+        $this->validated = true;
 
         return !$this->hasErrors();
     }
@@ -333,7 +333,7 @@ abstract class FormModel implements FormModelInterface
             unset($this->attributesErrors[$attribute]);
         }
 
-        $this->isValidate = false;
+        $this->validated = false;
     }
 
     private function getInflector(): Inflector
@@ -441,6 +441,6 @@ abstract class FormModel implements FormModelInterface
 
     public function isValidated(): bool
     {
-        return $this->isValidate;
+        return $this->validated;
     }
 }
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -439,7 +439,7 @@ abstract class FormModel implements FormModelInterface
         return [$attribute, $nested];
     }
 
-    public function isValidate(): bool
+    public function isValidated(): bool
     {
         return $this->isValidate;
     }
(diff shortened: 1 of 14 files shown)


terabytesoftw:
Empty form:

image

Validate form:

image

src/FormModelInterface.php
```diff
@@ -230,4 +230,11 @@ public function validate(): bool;
      * @param mixed $value value
      */
     public function setAttribute(string $name, $value): void;
+
+    /**
+     * This method allows to know if the validation was executed or not in the model.
+     *
+     * @return bool
+     */
+    public function isValidate(): bool;
```
roxblnfk, line 239:
```suggestion
    public function isValidated(): bool;
```
Maybe this is better?

terabytesoftw: done.

samdark:
I think it should be like this:

image

Note the spacing and alignment.

terabytesoftw:
> I think it should be like this:
> 
> image
> 
> Note the spacing and alignment.

If there is no problem that is fixed in the `yiisoft/app`, after this arrangement is merged, I will fix the spacing and alignment.

src/FormModel.php
```diff
@@ -32,6 +32,7 @@ abstract class FormModel implements FormModelInterface
     private array $attributesLabels;
     private array $attributesErrors = [];
     private ?Inflector $inflector = null;
+    private bool $isValidate = false;
```
samdark, line 35:
```suggestion
    private bool $validated = false;
```

src/FormModelInterface.php
```diff
@@ -230,4 +230,11 @@ public function validate(): bool;
      * @param mixed $value value
      */
     public function setAttribute(string $name, $value): void;
+
+    /**
+     * This method allows to know if the validation was executed or not in the model.
+     *
+     * @return bool
```
samdark, line 237:
```suggestion
     * @return bool If the model was validated.
```

src/Widget/Field.php
```diff
@@ -936,7 +936,11 @@ private function addSuccessCssClassToInput(): void
         if ($this->validationStateOn === 'input') {
             $attributeName = HtmlForm::getAttributeName($this->attribute);
 
-            if (!$this->data->hasErrors($attributeName)) {
+            if (
+                !$this->data->hasErrors($attributeName) &&
+                $this->data->isValidated() &&
+                !empty($this->data->getAttributeValue($attributeName))
```
samdark, line 942:
```suggestion
                $this->data->isValidated() &&
                !$this->data->hasErrors($attributeName) &&
                !empty($this->data->getAttributeValue($attributeName))
```

samdark: `!empty($this->data->getAttributeValue($attributeName))` condition looks odd. What about "should be empty" or "optional value" validation?

samdark (approved):

src/Widget/Field.php
terabytesoftw: not simply that it will have fields that do not have any validation and will be empty, so I think we should not put the has-success class, or by default if it does not have any validation put has-success on it, what do you think?

src/FormModelInterface.php
terabytesoftw: done.

src/FormModel.php
terabytesoftw: done.

samdark:
👍

Issue #65, opened 2020-12-30
[Feature Request] Adjust input following rules


Mister-42:
### What steps will reproduce the problem?
```php
    public function rules(): array
    {
        return [
            'amount' => [(new Number())->integer()->min(1)->max(250)],
        ];
    }
```

```php
echo $field->config($form, 'amount')->input('number');
```

Below is just 1 example, there are probably more validators that can be integrated into the fields.

### What is the expected result?
`<input type="number" id="form-amount" class="form-control" name="Form[amount]" max="250" min="1" placeholder="Amount">`

### What do you get instead?
`<input type="number" id="form-amount" class="form-control" name="Form[amount]" placeholder="Amount">`

### Additional info

| Q                | A
| ---------------- | ---
| Version          | dev
| PHP version      | 8.0.0
| Operating system | Debian 10

xepozz:
I've made a quick draft for that.
Example for [configuring form](https://github.com/yiisoft/form/blob/103434e07b60527b1c073c15ed416c0e4d1fb4dd/tests/Stub/HtmlOptionsForm.php#L25) and [final html](https://github.com/yiisoft/form/blob/103434e07b60527b1c073c15ed416c0e4d1fb4dd/tests/HtmlOptions/FieldInputTest.php#L20)

Mister-42:
Thank you!
This seem to be working with [`Field::widget`](https://github.com/yiisoft/form/blob/c6ccf9d03a0dcb0c3ae7bff170dac35e637ab85a/tests/HtmlOptions/FieldInputTest.php#L26), but not [`$field->config`](https://github.com/yiisoft/yii-demo/blob/master/views/contact/form.php#L34). Is this correct?

xepozz:
Could you please make a test case and show what isn't working?

Mister-42:
Not quickly, but you can use the yii-demo as a reference. For example on the contact page:

```
<div class="form-group field-contactform-name mb-3">
<label class="control-label form-label control-label required" for="contactform-name">Name</label>
<input type="text" id="contactform-name" class="form-control" name="ContactForm[name]" value="" placeholder="Name" data-cip-id="contactform-name">

<div class="help-block text-danger fst-italic"></div>
</div>
```

Required is missing from the input.

xepozz:
To enable html attributes you should wrap rules with html options provider decorators.
See https://github.com/yiisoft/yii-demo/pull/229

Mister-42:
[...]

PR #67, opened 2021-01-07
Transpose rule options to html attributes

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #65

### DIFF
--- a/src/HtmlOptions/EmailHtmlOptions.php
+++ b/src/HtmlOptions/EmailHtmlOptions.php
@@ -6,7 +6,7 @@ namespace Yiisoft\Form\HtmlOptions;
 
 use Yiisoft\Validator\Rule\Email;
 
-class EmailHtmlOptions implements HtmlOptionsProvider
+final class EmailHtmlOptions implements HtmlOptionsProvider
 {
     private Email $validator;
 
--- a/src/HtmlOptions/EmailHtmlOptions.php
+++ b/src/HtmlOptions/EmailHtmlOptions.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form\HtmlOptions;
+
+use Yiisoft\Validator\Rule\Email;
+
+class EmailHtmlOptions implements HtmlOptionsProvider
+{
+    private Email $validator;
+
+    public function __construct(Email $validator)
+    {
+        $this->validator = $validator;
+    }
+
+    public function getHtmlOptions(): array
+    {
+        return [
+            'type' => 'email',
+        ];
+    }
+}
--- a/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
+++ b/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
@@ -6,7 +6,7 @@ namespace Yiisoft\Form\HtmlOptions;
 
 use Yiisoft\Validator\Rule\MatchRegularExpression;
 
-class MatchRegularExpressionHtmlOptions implements HtmlOptionsProvider
+final class MatchRegularExpressionHtmlOptions implements HtmlOptionsProvider
 {
     private MatchRegularExpression $validator;
 
--- a/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
+++ b/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form\HtmlOptions;
+
+use Yiisoft\Validator\Rule\MatchRegularExpression;
+
+class MatchRegularExpressionHtmlOptions implements HtmlOptionsProvider
+{
+    private MatchRegularExpression $validator;
+
+    public function __construct(MatchRegularExpression $validator)
+    {
+        $this->validator = $validator;
+    }
+
+    public function getHtmlOptions(): array
+    {
+        $options = $this->validator->getOptions();
+        return [
+            'pattern' => $options['pattern'],
+        ];
+    }
+}
--- a/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
+++ b/src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Yiisoft\Form\HtmlOptions;
 
+use Yiisoft\Html\Html;
 use Yiisoft\Validator\Rule\MatchRegularExpression;
 
 final class MatchRegularExpressionHtmlOptions implements HtmlOptionsProvider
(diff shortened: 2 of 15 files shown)


Mister-42:
Ow cool 👍🏻 

From what I found [here](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation):
```
required: Specifies whether a form field needs to be filled in before the form can be submitted.
minlength and maxlength: Specifies the minimum and maximum length of textual data (strings)
min and max: Specifies the minimum and maximum values of numerical input types
type: Specifies whether the data needs to be a number, an email address, or some other specific preset type. 
pattern: Specifies a regular expression that defines a pattern the entered data needs to follow.
```

`required` has been taken care of already
`minlength and maxlength` - `HasLength`
`pattern` - `MatchRegularExpression`

xepozz:
Added.

terabytesoftw:
@xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.

terabytesoftw (approved):

xepozz:
> @xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.

Otherwise, there will be a names conflict

terabytesoftw:
> > @xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.
> 
> Otherwise, there will be a names conflict

Ok, let's merge it.

Mister-42:
> > > @xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.
> > 
> > 
> > Otherwise, there will be a names conflict
> 
> Ok, let's merge it.

Not yet. The changes are not applied when fields are created like `echo $field->config($form, 'amount')->input('number');`.

xepozz:
> > > > @xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.
> > > 
> > > 
> > > Otherwise, there will be a names conflict
> > 
> > 
> > Ok, let's merge it.
> 
> Not yet. The changes are not applied when fields are created like `echo $field->config($form, 'amount')->input('number');`.

Please provide full example or write test.

Mister-42:
> > > > > @xepozz It would be good to remove the prefix or suffix HtmlOptions, it is already in the directory and it sounds like repetitive.
> > > > 
> > > > 
> > > > Otherwise, there will be a names conflict
> > > 
> > > 
> > > Ok, let's merge it.
> > 
> > Not yet. The changes are not applied when fields are created like `echo $field->config($form, 'amount')->input('number');`.
> 
> Please provide full example or write test.

Full example is in the issue

Mister-42:
https://github.com/yiisoft/form/pull/68 hasfurther implications I see. Can't complain yet I guess

src/HtmlOptions/EmailHtmlOptions.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Yiisoft\Form\HtmlOptions;
+
+use Yiisoft\Validator\Rule\Email;
+
+final class EmailHtmlOptions implements HtmlOptionsProvider
+{
+    private Email $validator;
+
+    public function __construct(Email $validator)
```
samdark, line 13: Why do we need constructor? Validator instance isn't used anywhere.

src/HtmlOptions/MatchRegularExpressionHtmlOptions.php
```diff
@@ -0,0 +1,25 @@
...
+    public function __construct(MatchRegularExpression $validator)
+    {
+        $this->validator = $validator;
+    }
+
+    public function getHtmlOptions(): array
+    {
+        $options = $this->validator->getOptions();
+        return [
+            'pattern' => $options['pattern'],
```
samdark, line 22: The input pattern is PCRE that isn't suitable for HTML. Should be passed through `Html::normalizeRegexpPattern()`.

src/HtmlOptions/RequiredHtmlOptions.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Yiisoft\Form\HtmlOptions;
+
+use Yiisoft\Validator\Rule\Required;
+
+final class RequiredHtmlOptions implements HtmlOptionsProvider
+{
+    private Required $validator;
+
+    public function __construct(Required $validator)
```
samdark, line 13: Why do we need constructor? Validator instance isn't used anywhere.

src/Widget/Input.php
```diff
@@ -57,7 +58,13 @@ public function config(FormModelInterface $data, string $attribute, array $optio
         $new = clone $this;
         $new->data = $data;
         $new->attribute = $attribute;
-        $new->options = $options;
+        $rules = $data->rules()[$attribute] ?? [];
+        foreach ($rules as $rule) {
+            if ($rule instanceof HtmlOptionsProvider) {
```
samdark, line 63: Aren't rules meant to define validation rules? These would be always `Rule` instances.

tests/Stub/HtmlOptionsForm.php
```diff
@@ -0,0 +1,99 @@
...
+                $this->getHasLengthHtmlOptions(),
+                $this->getRequiredHtmlOptions(),
+                $this->getMatchRegularExpressionHtmlOptions(),
+            ],
+        ];
+    }
+
+    private function getMatchRegularExpressionHtmlOptions(): MatchRegularExpressionHtmlOptions
+    {
+        return new MatchRegularExpressionHtmlOptions(
```
samdark, line 63: How would that work for validation?

samdark (changes requested):

tests/Stub/HtmlOptionsForm.php
samdark: OK, I see it in https://github.com/yiisoft/form/pull/68.

Is it possible not to require me to wrap all my rules into `HtmlOptionsProvider`? That's too repetitive and we don't need config options there.

src/HtmlOptions/EmailHtmlOptions.php
xepozz: In [the next PR](https://github.com/yiisoft/form/pull/68) it will be [modified](https://github.com/yiisoft/form/blob/d5e4ed3be24e154cc1fff8ca463b4edb45843f2c/src/HtmlOptions/EmailHtmlOptions.php)

tests/Stub/HtmlOptionsForm.php
xepozz: This is how the decorator pattern works. Merging this functionality into validation rules would be the wrong decision.
The rule settings are used in `HtmlOptionsProvider`'s, so the solution with the decorator was chosen for a reason.

If you would like to suggest another solution, I am ready to consider it.

src/HtmlOptions/RequiredHtmlOptions.php
xepozz: The same as with `EmailHtmlOptions` above.

src/Widget/Input.php
xepozz: I don't understand what do you mean.

samdark: It doesn't work without next PR.

samdark:
Closed in favor of #68.

PR #69, opened 2021-01-24
Use stateless validator

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️

I hope this will be a temporary solution until the validator is able to validate any entities. In this case, you can try to omit the validate method.
Example how to validate form in controller.
```php
function action(ValidatorInterface $validator, $request)
{
    $form = new Form();
    $form->load($request);
    $form->validate($validator);

    return $this->render('view', ['form' => $form]);
}
```


src/FormModel.php
```diff
@@ -214,16 +209,14 @@ public function setAttribute(string $name, $value): void
         }
     }
 
-    public function validate(): bool
+    public function validate(ValidatorInterface $validator): bool
```
samdark, line 212: Since validator is stateless, can it be dependency passed via `__construct()`?

xepozz: Yes, but why?
Creating form model with `new` is more handy then with factory/mixing not real dependencies with real in a constructor.
```php
$form = new AuthForm($param1, $param2);
```
vs

```php
$form = new AuthForm($param1, $param2, $validator);
```

xepozz: Also I hope `validate` will be omitted and forms will validates with validator like this: `$validator->validate($form)`.

samdark: Currently forms are obtained via type-hinting anyway:

https://github.com/yiisoft/yii-demo/blob/master/src/Contact/ContactController.php#L35

xepozz: Getting forms from DIC is wrong.

samdark: Discussed. It's better to be as in this PR:

1. Storing form instances in container isn't a good idea anyway.
2. `$form->validate($validator->withFormatter($myFormatter))` is better when validator is available.

samdark:
👍

PR #71, opened 2021-02-08
Start using external validation instead of internal

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Depends on  | https://github.com/yiisoft/validator/pull/115

After this PR we can validate forms like the following: `$validator->validate($form)`.
Remind that now we need to pass the validator to the form over form's method `validate($validator)`.

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -10,8 +10,10 @@ use ReflectionClass;
 use Yiisoft\Form\HtmlOptions\HtmlOptionsProvider;
 use Yiisoft\Strings\Inflector;
 use Yiisoft\Strings\StringHelper;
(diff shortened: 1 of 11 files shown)


src/FormModel.php
```diff
@@ -244,6 +236,11 @@ public function rules(): array
         return [];
     }
 
+    public function getRules(): array
+    {
+        return $this->rules();
```
xepozz, line 241: I want to rename form's `rules` to `getRules` instead of validator's `getRules` to `rules` :)

samdark: ???

samdark (approved):

PR #75, opened 2021-02-18
Adopt to yiisoft/html PR #52.

vjik:
Adopt to https://github.com/yiisoft/html/pull/52

### DIFF
--- a/src/Widget/ListInput.php
+++ b/src/Widget/ListInput.php
@@ -4,10 +4,14 @@ declare(strict_types=1);
 
 namespace Yiisoft\Form\Widget;
 
+use Closure;
+use RuntimeException;
 use Yiisoft\Arrays\ArrayHelper;
 use Yiisoft\Form\FormModelInterface;
 use Yiisoft\Form\Helper\HtmlForm;
 use Yiisoft\Html\Html;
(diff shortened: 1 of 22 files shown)


src/Widget/ListInput.php
```diff
@@ -32,17 +41,120 @@ final class ListInput extends Widget
     public function run(): string
```
terabytesoftw, line 41: Because I pass the `renderSelectOptions(`) method from `HTML` to the `Widget`, that's wrong, it's not the widget's job, it's passing all the `HTML` logic to the widgets.

vjik: I think this temporary. `ListInput` need refactoring and make him consistent with the `yiisoft/html` package.

terabytesoftw: The problem is not `yiisoft/form`, the problem is `yiisoft/html`, because if you are refactoring `html` you don't fix it instead of transferring the problem to the `widget`.

vjik: ListInput need split to several class and make receiving incoming data as well as yiisoft/html.

samdark (approved):

PR #78, opened 2021-02-25
Extract attributes

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

This PR is concept to avoid attribute* properties in a base form model.
Also there are some improvements and bugfixes.

If we can abstract attribute's characteristics from `FormModelInterface` (getLabel, getHint, getEtc) this package will be able to use in api-form (instead of request-model), but I don't sure about that.

At the moment NestedAttributes's tests are failed. I will fix it If this PR is ok.

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -235,11 +235,6 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
         $this->validated = true;
     }
 
-    public function addError(string $attribute, string $error): void
-    {
-        $this->attributesErrors[$attribute][] = $error;
-    }
-
     public function getRules(): array
     {
         return [];
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -86,8 +86,8 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
 
     public function getAttributeLabel(string $attribute): string
     {
-        if (array_key_exists($attribute, $this->attributes)) {
-            return $this->attributes[$attribute]->getLabel();
+        if ($this->hasAttribute($attribute)) {
+            return $this->getAttribute($attribute)->getLabel();
         }
 
         [$attribute, $nested] = $this->getNestedAttribute($attribute);
@@ -114,6 +114,14 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
         return substr($className, 1);
     }
 
+    private function getAttribute(string $attribute): FormAttribute
+    {
+        if (!$this->hasAttribute($attribute)) {
+            throw new InvalidArgumentException('Attribute ' . $attribute . ' does not exist.');
+        }
+        return $this->attributes[$attribute];
+    }
+
     public function hasAttribute(string $attribute): bool
     {
         return array_key_exists($attribute, $this->attributes);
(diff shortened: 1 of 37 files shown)


src/FormModel.php
```diff
@@ -227,36 +243,21 @@ public function setAttribute(string $name, $value): void
...
-            if ($result->isValid() === false) {
-                $this->addErrors([$attribute => $result->getErrors()]);
+            if ($result->isValid() === false && ($errors = $result->getErrors()) !== []) {
+                $this->attributes[$attribute]->setErrors($errors);
             }
         }
         $this->validated = true;
     }
 
-    public function addError(string $attribute, string $error): void
```
vjik, line 238: Are you sure what doesn't exists use cases for add error to model from user code?

src/FormModel.php
```diff
@@ -271,36 +272,44 @@ private function collectAttributes(): array
...
         return $attributes;
     }
 
     private function clearErrors(?string $attribute = null): void
     {
-        if ($attribute === null) {
-            $this->attributesErrors = [];
-        } else {
-            unset($this->attributesErrors[$attribute]);
+        if ($attribute !== null) {
```
vjik, line 307: Method `clearErrors` always called without attribute, we can remove this code.

vjik (changes requested):
It's concept is good for me :)

src/FormModel.php
```diff
@@ -271,36 +272,44 @@ private function collectAttributes(): array
...
-                    $property->getName(),
+                    $propertyName,
                     $property->getDeclaringClass()->getName(),
                 ));
             }
 
-            $attributes[$property->getName()] = $type->getName();
+            $attributes[$propertyName] = new FormAttribute(
+                $propertyName,
+                $propertyType->getName(),
```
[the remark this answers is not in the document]
xepozz: Why?

src/FormModel.php
```diff
@@ -271,36 +272,44 @@ private function collectAttributes(): array
...
+                $attributeHints[$propertyName] ?? $this->getAttributeHint($propertyName),
+                $attributeLabels[$propertyName] ?? $this->getAttributeLabel($propertyName),
```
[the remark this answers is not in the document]
xepozz: `getAttributeLabel` and other will be removed later. Firstly we need to accept this concept.

src/FormModel.php
xepozz: It is a violation of encapsulation so I don't want to have this method.
But we can add it back when the use case will appear.

src/FormModel.php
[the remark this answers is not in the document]
vjik: This will allow to do support nulled attributes in the future. For example, `?int $count`. On set empty string to attribute `count` value will be `null`.

src/FormModel.php
samdark: This one is questionable. There are use cases for adding error without having validator. For example, if we handle a form via external API and don't want to call two methods of this API but still want to display errors in the form per field.

samdark:
Overall direction is alright despite removal of `addError()` is questionable. Let's finish this pull request.

PR #92, opened 2021-10-06
Fix Checkbox::value().

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


terabytesoftw:
@vjik You could check it, something is not right, if I send two consecutive checkboxes, the first one gives it as true value, the second, even if it is marked, gives it false.

PR #146, opened 2021-11-03
Add docs `Label` widget.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


docs/label.md
```diff
@@ -0,0 +1,120 @@
...
+    <hr class="mt-3">
+    <div>
+        <input type="submit" id="submit-186948574010001" class="button is-block is-info is-fullwidth" name="submit-186948574010001" value="Save">
+    </div>
+</form>
+```
+
+### Custom label text
+
+To configure a custom label text, we simply change it in the widget, taking the previous example the code would be the following:
```
samdark, line 72:
```suggestion
To configure a custom label text, you can specify it when the widget is used: 
```

docs/label.md
```diff
@@ -0,0 +1,120 @@
...
+</form>
+```
+
+### `Label` methods:
+
+Method | Description | Default
+-------|-------------|---------
+`config(FormModelInterface $formModel, string $attribute, array $attributes = [])` | Configures the widget. |
+`encode(bool $value)` | Whether to encode the error message. | `true`
+`for(?string $value)` | The id of a labelable form-related element in the same document as the tag label element. | `''`
```
samdark, line 119:
```suggestion
`for(?string $value)` | The ID of a form element to associate label with. | `''`
```

src/Widget/Label.php
```diff
@@ -4,39 +4,78 @@
...
  * Generates a label tag for the given form attribute.
  *
  * @link https://www.w3.org/TR/2012/WD-html-markup-20120329/label.html
+ *
+ * @psalm-suppress MissingConstructor
  */
 final class Label extends Widget
 {
-    use ModelAttributes;
+    private array $attributes = [];
```
samdark, line 21:
```suggestion
    private array $tagAttributes = [];
```

src/Widget/Label.php
```diff
@@ -4,39 +4,78 @@
...
+    private FormModelInterface $formModel;
+    private ?string $label = '';
 
-    private string $label = '';
+    /**
+     * Set form interface, attribute name and attributes, and attributes for the widget.
+     *
+     * @param FormModelInterface $formModel Form.
+     * @param string $attribute Form model property this widget is rendered for.
+     * @param array $attributes The HTML attributes for the widget container tag.
```
samdark, line 33:
```suggestion
     * Specify a form, its attribute and a list HTML attributes for the label generated.
     *
     * @param FormModelInterface $formModel Form.
     * @param string $attribute Form model property this widget is rendered for.
     * @param array $attributes The HTML attributes for the widget container tag.
```

samdark (changes requested):

src/Widget/Label.php
terabytesoftw: https://github.com/terabytesoftw/yii-form/blob/883d98d085c527eb8387f1f2db126029ba0d4c0b/src/Widget/Label.php#L39 those are the global attributes of the widget.

terabytesoftw: done.

samdark (approved):

PR #151, opened 2021-11-09
Allow null value for input id.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Widget/Attribute/FieldAttributes.php
+++ b/src/Widget/Attribute/FieldAttributes.php
@@ -225,7 +225,7 @@ trait FieldAttributes
      *
      * @return string
      */
-    private function getId(): string
+    private function generateId(): string
     {
         $new = clone $this;
 
@@ -246,7 +246,7 @@ trait FieldAttributes
         $attributeName = HtmlForm::getAttributeName($new->getFormModel(), $new->attribute);
 
         if ($new->ariaDescribedBy === true) {
-            $attributes['aria-describedby'] = $new->getId();
+            $attributes['aria-describedby'] = $new->generateId();
         }
 
         if ($new->inputClass !== '') {
--- a/src/Widget/Attribute/FieldAttributes.php
+++ b/src/Widget/Attribute/FieldAttributes.php
@@ -225,14 +225,12 @@ trait FieldAttributes
      *
      * @return string
      */
-    private function generateId(): string
+    private function getId(): string
     {
-        $new = clone $this;
-
         /** @var string */
-        $id = $new->attributes['id'] ?? $new->id;
+        $id = $this->attributes['id'] ?? $this->id;
 
-        return $id === '' ? HtmlForm::getInputId($new->getFormModel(), $new->attribute) : $id;
+        return $id === '' ? HtmlForm::getInputId($this->getFormModel(), $this->attribute) : $id;
     }
 
     private function setInputAttributes(array $attributes): array
@@ -246,7 +244,7 @@ trait FieldAttributes
         $attributeName = HtmlForm::getAttributeName($new->getFormModel(), $new->attribute);
 
         if ($new->ariaDescribedBy === true) {
-            $attributes['aria-describedby'] = $new->generateId();
+            $attributes['aria-describedby'] = $new->getId();
         }
 
         if ($new->inputClass !== '') {
--- a/src/Widget/Attribute/ModelAttributes.php
+++ b/src/Widget/Attribute/ModelAttributes.php
@@ -54,7 +54,7 @@ trait ModelAttributes
     /**
      * Set the ID of the widget.
      *
-     * @param string|null $value The ID of the widget. if null the attribute will be removed.
+     * @param string|null $value The ID of the widget. if `null` the attribute will be removed.
      *
      * @return static
      *
--- a/src/Widget/Attribute/ModelAttributes.php
+++ b/src/Widget/Attribute/ModelAttributes.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Yiisoft\Form\Widget\Attribute;
 
 use InvalidArgumentException;
+use Yiisoft\Arrays\ArrayHelper;
 use Yiisoft\Form\FormModelInterface;
 use Yiisoft\Form\Helper\HtmlForm;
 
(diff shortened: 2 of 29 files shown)


src/Widget/Attribute/ModelAttributes.php
```diff
@@ -13,7 +14,7 @@ trait ModelAttributes
     private array $attributes = [];
     private string $attribute = '';
     private string $charset = 'UTF-8';
-    private string $id = '';
+    private ?string $id = '';
```
vjik, line 17:
```suggestion
    private ?string $id = null;
```

src/Widget/Attribute/ModelAttributes.php
```diff
@@ -53,13 +54,13 @@ public function config(FormModelInterface $formModel, string $attribute, array $
     /**
      * Set the ID of the widget.
      *
-     * @param string $value
+     * @param string|null $value The ID of the widget. if null the attribute will be removed.
```
vjik, line 57:
```suggestion
     * @param string|null $value The ID of the widget. if `null` the attribute will be removed.
```

src/Widget/Attribute/ModelAttributes.php
```diff
@@ -80,12 +81,12 @@ protected function getFormModel(): FormModelInterface
...
      * @return string
      */
-    protected function getId(): string
+    protected function getId(): ?string
     {
         $new = clone $this;
 
         /** @var string */
-        $id = $new->attributes['id'] ?? $new->id;
+        $id = ArrayHelper::remove($new->attributes, 'id', $new->id);
```
vjik, line 89: Why need remove `id` from attributes?

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: Because we are going to eliminate the id by default, it is necessary that the widgets have an id.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: done.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: Because, i am using a variable that is easier to validate than an array with `isset()`, also in all widgets i use the valid method `->id()`, instead of passing it through the attributes, there is more control.

src/Widget/Attribute/ModelAttributes.php
vjik: All widgets need id?

src/Widget/Attribute/ModelAttributes.php
vjik: Confuse `clone` and modifing clonned object in getter.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: If they all have id, some use `HtmlForm::getInputId()`, which is generated by the model, others use `Html::generateId()`.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: Why is it confusing?

terabytesoftw: fixed.

src/Widget/Attribute/FieldAttributes.php
```diff
@@ -225,7 +225,7 @@ private function addValidatorAttributeHtml(
      *
      * @return string
      */
-    private function getId(): string
+    private function generateId(): string
     {
         $new = clone $this;
```
samdark, line 230: This method doesn't mutate the state. Nothing is written to `$this` so clone isn't needed.

src/Widget/Attribute/FieldAttributes.php
```diff
@@ -225,7 +225,7 @@ private function addValidatorAttributeHtml(
      *
      * @return string
      */
-    private function getId(): string
+    private function generateId(): string
```
samdark, line 228: I think `getId()` was a bit better since you're not generating ID every time but only if it's missing.

src/Widget/Attribute/ModelAttributes.php
```diff
@@ -53,13 +54,13 @@ public function config(FormModelInterface $formModel, string $attribute, array $
     /**
      * Set the ID of the widget.
      *
-     * @param string $value
+     * @param string|null $value The ID of the widget. if `null` the attribute will be removed.
```
samdark, line 57:
```suggestion
     * @param string|null $value The ID of the widget. `null` means that the attribute will be removed.
```

src/Widget/Attribute/ModelAttributes.php
```diff
@@ -76,16 +77,16 @@ protected function getFormModel(): FormModelInterface
...
 
     /**
-     * Return the input id.
+     * Generates a unique ID for the attribute.
      *
-     * @return string
+     * @return string|null
      */
-    protected function getId(): string
+    protected function generateId(): ?string
```
samdark, line 84: Old name was OK.

src/Widget/Attribute/WithoutModelAttribute.php
```diff
@@ -92,13 +92,18 @@ public function value(string $value): self
         return $new;
     }
 
-    protected function getId(): string
+    /**
+     * Generates a unique ID for the attribute.
+     *
+     * @return string
+     */
+    protected function generateId(): string
```
samdark, line 100: Old name was OK.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: done.

src/Widget/Attribute/FieldAttributes.php
terabytesoftw: done.

src/Widget/Attribute/FieldAttributes.php
terabytesoftw: done.

src/Widget/Attribute/ModelAttributes.php
terabytesoftw: done.

src/Widget/Attribute/WithoutModelAttribute.php
terabytesoftw: done.

samdark (approved):

samdark:
👍

PR #157, opened 2021-11-14
Separate templates for each widget

Gerych1984:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

Good day. After widget has own config (widgets.php) became possible set/change base config for all widgets in all pages - and it's cool. But today we can't set template in field for current widget. I used that trick

```
Field::class => [
    'template()' => [
        '{label}<div class="col-sm-10">{input}{error}</div>'
    ],
    ****
],

'ButtonField' => [
    'class' => Field::class,
    'template()' => [
        '<div class="offset-sm-2 col-sm-10 text-center text-sm-start">{input}</div>'
    ],
    ****
],

// In some view
$buttonField = Widget::widget('ButtonField');
```
It's works, but uncomfortable (no autocopletetion in IDE and etc).

So that PR maybe resolve that problem. With this we can write that base config:

```
Field::class => [
    'template()' => [
        '{label}<div class="col-sm-10">{input}{error}</div>'
    ],
    'widgetTemplates()' => [
           [
               SubmitButton::class => '<div class="offset-sm-2 col-sm-10 text-center text-sm-start">{input}</div>',
               OtherFormWidget::class => '<div class="custom-class">{input}</div>'
           ]
     ]
],  

```
If that wil be interesting - i will be think about `inputClass`, `containerClass`. Thanks

Little remark/issue - maybe separate `submitButton`/`submitInput`, `resetButton`/`resetInput` methods? First - render more flexible button tag. Second - submit input tag (as current submitButton method)

### DIFF
--- a/src/Widget/Attribute/FieldAttributes.php
+++ b/src/Widget/Attribute/FieldAttributes.php
@@ -31,6 +31,7 @@ trait FieldAttributes
     private string $validClass = '';
     private array $parts = [];
     private string $template = "{label}\n{input}\n{hint}\n{error}";
+    private array $widgetTemplates = [];
     private string $validationStateOn = 'input';
     private ?FormModelInterface $formModel = null;
 
@@ -151,6 +152,48 @@ trait FieldAttributes
     {
         $new = clone $this;
         $new->template = $value;
+        return $new;
+    }
+
+    /**
+     * Set layout templates for each widget
+     *
+     * @param array $templates
+     * @param bool $replace
+     *
+     * @return static
+     */
+    public function widgetTemplates(array $templates, bool $replace = false): self
+    {
+        $new = clone $this;
+
+        if ($replace) {
+            $new->widgetTemplates = $templates;
+        } else {
+            $new->widgetTemplates = array_merge($new->widgetTemplates, $templates);
+        }
+
+        return $new;
+    }
+
+    /**
+     * Set layout template for current widget
+     *
+     * @param string $widgetClass
+     * @param string|null $template
+     *
+     * @return static
+     */
+    public function widgetTemplate(string $widgetClass, ?string $template): self
+    {
+        $new = clone $this;
+
+        if (empty($template)) {
+            unset($new->widgetTemplates[$widgetClass]);
+        } else {
+            $new->widgetTemplates[$widgetClass] = $template;
+        }
+
         return $new;
     }
 
--- a/src/Widget/Field.php
+++ b/src/Widget/Field.php
@@ -85,6 +85,7 @@ final class Field extends Widget
     public function checkbox(array $attributes = [], bool $enclosedByLabel = true): self
     {
         $new = clone $this;
+        $new->template = $this->widgetTemplates[Checkbox::class] ?? $this->template;
         $checkbox = Checkbox::widget();
         $attributes['type'] = self::TYPE_CHECKBOX;
         $attributes = $new->setInputAttributes($attributes);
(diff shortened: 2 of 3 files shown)


terabytesoftw:
The widgets are just the generation of the tag, this is a bad idea if you need to change the template it is simple:

```php
<?= $field->containerClass('d-grid gap-2 form-floating')->submitButton(
    [
        'class' => 'btn btn-primary btn-lg mt-3',
        'id' => 'login-button',
        'tabindex' => ++$tab,
         'value' => $translator->translate('Log in', [], 'user-view'),
        ]
) ?>
```

```html
<div class="d-grid gap-2 form-floating">
    <input type="submit" id="register-button" class="btn btn-primary btn-lg mt-3" name="register-button" value="Register" tabindex="4">
</div>
```

Gerych1984:
> The widgets are just the generation of the tag, this is a bad idea if you need to change containerClass:
> 
> ```
> <?= $field->containerClass('d-grid gap-2 form-floating')->submitButton(
>     [
>         'class' => 'btn btn-primary btn-lg mt-3',
>         'id' => 'login-button',
>         'tabindex' => ++$tab,
>          'value' => $translator->translate('Log in', [], 'user-view'),
>         ]
> ) ?>
> ```
> 
> ```
> <div class="d-grid gap-2 form-floating">
>     <input type="submit" id="register-button" class="btn btn-primary btn-lg mt-3" name="register-button" value="Register" tabindex="4">
> </div>
> ```

I meant separate containerClass in base config. For example - all containerClass equals bs5 "row mb-3" (from bootstrap demo page). But for `submitButton` equals "row", because in bottom form (where submit present in many times) not needed extended margin bottom. And if user or customer make change this - we can change it in one place (widgets.php) instead of every "some view.php"

terabytesoftw:
> > The widgets are just the generation of the tag, this is a bad idea if you need to change containerClass:
> > ```
> > <?= $field->containerClass('d-grid gap-2 form-floating')->submitButton(
> >     [
> >         'class' => 'btn btn-primary btn-lg mt-3',
> >         'id' => 'login-button',
> >         'tabindex' => ++$tab,
> >          'value' => $translator->translate('Log in', [], 'user-view'),
> >         ]
> > ) ?>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > <div class="d-grid gap-2 form-floating">
> >     <input type="submit" id="register-button" class="btn btn-primary btn-lg mt-3" name="register-button" value="Register" tabindex="4">
> > </div>
> > ```
> 
> I meant separate containerClass in base config. For example - all containerClass equals bs5 "row mb-3" (from bootstrap demo page). But for equals "row", because in bottom form (where submit present in many times) not needed extended margin bottom. And if user or customer make change this - we can change it in one place (widgets.php) instead of every "some view.php"`submitButton`

I understand what you say about configuring the widget from the factory so as not to change it in all views, i will think of a solution, but the widget should only return the `tag`, otherwise it can bring confusion.

terabytesoftw:
If you use `CommonParameters` in yii-view for the `Fied: widget(), which prevents you from passing a `SubmitButton::class` custom all the views:

```php
public function getCommonParameters(): array
{
     return [
         'submit' => $field->containerClass('d-grid gap-2 form-floating')->submitButton()->render(),
     ]
}
```

Gerych1984:
> If you use `CommonParameters` in yii-view for the `Fied:widget()`, which prevents you from passing a `SubmitButton::class` custom all the views:
> 
> ```
> public function getCommonParameters(): array
> {
>      return [
>          'submit' => $field->containerClass('d-grid gap-2 form-floating')->submitButton()->render(),
>      ]
> }
> ```

In this case (with common parameters) I not understand  - why needed base widgets.php?

terabytesoftw:
closed https://github.com/yiisoft/form/pull/160

PR #159, opened 2021-11-14
Big refactoring of field widgets

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -


samdark:
`Field::something()` is hardcoded as well as `FieldFactory`. For any non-standard widget end user will need to write a lof of code to achieve simiar functionality. So, the solution provided isn't universal and won't allow community to create their own field types with minimal effort.

For example, a task of adding "a field to select cloth size range" doesn't look easy to me in this PR but it should be.

I like that fields are divided into parts such as field itself, label etc. That would allow to use use these within HTML markup if needed.

First error change is OK as well but should have been separate as it's not related.

vjik:
Closed in favor https://github.com/yiisoft/form/pull/192

PR #169, opened 2022-01-22
Allow the use of the `ButtonGroup::class` container.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


terabytesoftw:
@Gerych1984 suggested using the `ButtonGroup::class` container which by default was disabled, this is a good suggestion for group buttons.

samdark (approved):

PR #170, opened 2022-01-22
Add method `onlyAttributes()` to `ErrorSummary::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

Example: exclude attribute `name` in `ErrorSummary::class`.
```php
<?= Form::widget()->action('widgets')->csrf($csrf)->begin(); ?>
    <?= Text::widget()->for($formModel, 'name') ?>
    <?= Text::widget()->for($formModel, 'email') ?>
    <?= ErrorSummary::widget()->attributes(['class' => 'has-text-danger'])->excludeAttributes('name')->model($formModel) ?>
    <hr class="mt-3">
    <?= Field::widget()->submitButton(['class' => 'button is-block is-info is-fullwidth', 'value' => 'Save']); ?>
<?= Form::end(); ?>
````

Code generate:
```HTML
<form action="widgets" method="POST" _csrf="vWob09HzPhcTDuhOHVU71VJQAAymEm2Hysn_8QN1Y8qOXWK9tKd9J2RYsDd8DVqNJGcxduVIAvOMrq3FSjQpoQ==">
    <input type="hidden" name="_csrf" value="vWob09HzPhcTDuhOHVU71VJQAAymEm2Hysn_8QN1Y8qOXWK9tKd9J2RYsDd8DVqNJGcxduVIAvOMrq3FSjQpoQ==">
    <input type="text" id="testform-name" name="TestForm[name]" minlength="4">
    <input type="text" id="testform-email" name="TestForm[email]">
    <div class="has-text-danger">
        <p>Please fix the following errors:</p>
        <ul>
            <li>This value is not a valid email address.</li>
        </ul>
    </div>
    <hr class="mt-3">
    <input type="submit" id="w1-submit" class="button is-block is-info is-fullwidth" name="w1-submit" value="Save">
</form>
```

### DIFF
--- a/docs/error-summary.md
+++ b/docs/error-summary.md
@@ -68,11 +68,11 @@ use Yiisoft\Form\Widget\Text;
 ?>
 
 <?= Form::widget()->action('widgets')->csrf($csrf)->begin(); ?>
-    <?= Text::widget()->config($formModel, 'name') ?>
-    <?= Text::widget()->config($formModel, 'email') ?>
+    <?= Text::widget()->for($formModel, 'name') ?>
+    <?= Text::widget()->for($formModel, 'email') ?>
     <?= ErrorSummary::widget()->model($formModel) ?>
     <hr class="mt-3">
-    <?= Field::widget()->submitButton(['class' => 'button is-block is-info is-fullwidth', 'value' => 'Save']); ?>
+    <?= SubmitButton::widget()->attributes(['class' => 'button is-block is-info is-fullwidth'])->value('Save') ?>
 <?= Form::end(); ?>
 ```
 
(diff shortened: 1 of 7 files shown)


samdark:
The issue's use-case is better to be done with `onlyAttributes()` instead of `excludeAttributes()`. At least resulting template will be way more readable.

terabytesoftw:
> The issue's use-case is better to be done with `onlyAttributes()` instead of `excludeAttributes()`. At least resulting template will be way more readable.

done.

docs/error-summary.md
```diff
@@ -132,53 +128,83 @@ use Yiisoft\Form\Widget\Text;
...
-<form action="widgets" method="POST" _csrf="PwqhEiBBULqjUwjVQFBIGDt2iral9LmaJNOkfEJe-BcMPdh8RRUTitQFUKwhCClATUG7zOau1u5itPZICx-yfA==">
-    <input type="hidden" name="_csrf" value="PwqhEiBBULqjUwjVQFBIGDt2iral9LmaJNOkfEJe-BcMPdh8RRUTitQFUKwhCClATUG7zOau1u5itPZICx-yfA==">
-    <input type="text" id="testform-name" name="TestForm[name]">
+<form action="widgets" method="POST" _csrf="vWob09HzPhcTDuhOHVU71VJQAAymEm2Hysn_8QN1Y8qOXWK9tKd9J2RYsDd8DVqNJGcxduVIAvOMrq3FSjQpoQ==">
+    <input type="hidden" name="_csrf" value="vWob09HzPhcTDuhOHVU71VJQAAymEm2Hysn_8QN1Y8qOXWK9tKd9J2RYsDd8DVqNJGcxduVIAvOMrq3FSjQpoQ==">
+    <input type="text" id="testform-name" name="TestForm[name]" minlength="4">
     <input type="text" id="testform-email" name="TestForm[email]">
     <div class="has-text-danger">
-        Custom Header:
+        <p>Please fix the following errors:</p>
```
vjik, line 201: Seems, here should output errors for "name", but not displayed error for "email".

terabytesoftw: thks, done.

src/Widget/ErrorSummary.php
```diff
@@ -178,8 +193,10 @@ private function collectErrors(): array
         $errors = HtmlFormErrors::getErrorSummaryFirstErrors($this->formModel);
         $errorMessages = [];
 
-        if ($this->showAllErrors) {
+        if ($this->showAllErrors && $this->onlyAttributes === []) {
```
vjik, line 196: This code incorrect work if `showAllErrors` is `true` and `onlyAttributes` is non-empty(). We recevied first errors instead all errors for defined attributes.

vjik (changes requested):

src/Widget/ErrorSummary.php
terabytesoftw: The code is not incorrect, when we use `showAllErrors`, as its name indicates, all errors must be shown, they come in a list without a key that cannot be filtered, that is why it is called showAllErrors, it doesn't make sense to show all errors with filtering, and I told you about that @samdark

samdark: `firstErrorsOfAttributes(['a', 'b', 'c'])`

vjik: > `showAllErrors` is `true` and `onlyAttributes` is non-empty().

In case `showAllErrors` is true and `onlyAttributes` is non-empty() we should received all errors for defined attrbites. In current code returned only first errors. it's not right.

terabytesoftw: @vjik If you are right, I fixed it, the correction is correct, also change the name of the method to `firstErrorsOfAttributes(['a', 'b', 'c'])`, to avoid confusion, since if showAllErrors is true, they should be shown all the mistakes.

vjik: In `showAllErrors` part "all" meat is not all attributes, it's all errors for attributes.
`onlyAttrbiutes` — filter attributes.
Names are good, need only small change behavior.

terabytesoftw: > In case is true and is non-empty we should received all errors for defined attrbites. In current code returned only first errors. it's not right.`showAllErrors``onlyAttributes`
> 
> I think and should be two unrelated options.`showAllErrors``onlyAttributes`
> 
> `showAllErrors` — show all errors for each attribute or first errors only.
> 
> `onlyAttrbiutes` — show errors only for defined attribtes.

I think that `showAllErrors()` should show all errors regardless if it is a first error or not, so filtering should only be used when `showAllErrors()` is `false`.

vjik: > > In case is true and is non-empty we should received all errors for defined attrbites. In current code returned only first errors. it's not right.`` showAllErrors``onlyAttributes ``
> > I think and should be two unrelated options.`` showAllErrors``onlyAttributes ``
> > `showAllErrors` — show all errors for each attribute or first errors only.
> > `onlyAttrbiutes` — show errors only for defined attribtes.
> 
> I think that `showAllErrors()` should show all errors regardless if it is a first error or not, so filtering should only be used when `showAllErrors()` is `false`.

This is don't allow get all errors for defined attributes only.

terabytesoftw: > In `showAllErrors` part "all" mean is not all attributes, it's all errors for attributes. `onlyAttrbiutes` — filter attributes. Names are good, need only small change behavior.

done.

terabytesoftw: Now `only Attributes()` will show errors only for the selected attributes, be they `showAllErrors()` or `firtsErrors()`.

vjik (approved):

samdark (approved):

PR #172, opened 2022-01-26
Remove unnecesary clone in `Form::class`, add `fieldset::class` widget.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Widget/Form.php
+++ b/src/Widget/Form.php
@@ -7,6 +7,7 @@ namespace Yiisoft\Form\Widget;
 use InvalidArgumentException;
 use Stringable;
 use Yiisoft\Html\Html;
+use Yiisoft\Html\Tag\CustomTag;
 use Yiisoft\Http\Method;
 use Yiisoft\Widget\Widget;
 
@@ -27,6 +28,10 @@ final class Form extends Widget
     private array $attributes = [];
     private string $csrfName = '';
     private string $csrfToken = '';
+    private bool $fieldset = false;
+    private array $fieldsetAttributes = [];
+    private ?string $legend = null;
+    private array $legendAttributes = [];
     private string $id = '';
     private string $method = Method::POST;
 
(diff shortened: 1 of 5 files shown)


src/Widget/Form.php
```diff
@@ -197,6 +211,38 @@ public function enctype(string $value): self
...
+    /**
+     * The <fieldset> HTML element is used to group several controls as well as labels (<label>) within a web form.
+     *
+     * @param bool $value whether the fieldset is enabled or disabled.
+     *
+     * @return static
+     *
+     * @link https://html.spec.whatwg.org/multipage/form-elements.html#the-fieldset-element
+     */
+    public function fieldset(bool $value): self
```
samdark, line 223: I don't think there's a point adding fieldset as a boolean flag that just wraps all the content. See https://jsfiddle.net/6fz3nvLr/ for typical usage. It is multiple fieldsets per form.

src/Widget/Form.php
```diff
@@ -211,6 +257,38 @@ public function id(string $value): self
...
+    /**
+     * The <legend> HTML element represents a caption for the content of its parent <fieldset>.
+     *
+     * @param string|null $value whether the legend is enabled or disabled.
+     *
+     * @return static
+     *
+     * @link https://html.spec.whatwg.org/multipage/form-elements.html#the-legend-element
+     */
+    public function legend(?string $value): self
```
samdark, line 269: Same as for fieldset.

samdark (changes requested):

src/Widget/Form.php
terabytesoftw: yes, if it makes sense.

src/Widget/Form.php
terabytesoftw: yes, if it makes sense.

src/Widget/Form.php
terabytesoftw: fixed.

src/Widget/Form.php
terabytesoftw: fixed.

terabytesoftw: fixed.

terabytesoftw:
Add tests in FieldsetTest:

image

src/Widget/Form.php
terabytesoftw: done.

samdark (approved):

Issue #101, opened 2021-10-14
Type casting at load, why?


ruslansp:
If you do not specify the type in the form, then we will get an exception, if you specify the type, for example, a number, and a string comes from the user, then with $form->load()  we lose raw data.
We lose meaning in validation, because before validation, there is a type conversion

samdark:
@ruslansp do you have a valid real example for such situation?

ruslansp:
Users do not indicate anything. Most often these are typos, an example in the field you need to enter a number, enter "123m" number and randomly a letter.
In any case, it is fundamentally incorrect that first the type is given and then the validation takes place.

PR #175, opened 2022-02-04
Remove cast before validation.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

Now `getAttributeValue()` returns the raw data if it is loaded by the load method, which allows validating the data in its raw state.

https://github.com/yiisoft/validator/blob/d8d7b602d44f21f6fac1fda4164771238881bc10/src/Validator.php#L48

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -25,10 +25,12 @@ use function strpos;
  */
 abstract class FormModel implements FormModelInterface, PostValidationHookInterface, RulesProviderInterface
 {
-    private string $formErrorsClass = FormErrors::class;
     private array $attributes;
+    private string $formErrorsClass = FormErrors::class;
     private FormErrorsInterface $formErrors;
     private ?Inflector $inflector = null;
+    /** @psalm-var array<string, string|array> */
+    private array $rawData = [];
     private bool $validated = false;
 
     public function __construct()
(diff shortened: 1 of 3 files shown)


xepozz:
Now we have one more public method that does almost the same. What's the problem so you want to solve?

terabytesoftw:
> Now we have one more public method that does almost the same. What's the problem so you want to solve?

https://github.com/yiisoft/form/issues/101#issuecomment-946844022

xepozz:
Then I suggest to not to do method public and encapsulate login with validation inside form (forms + validator under the hood)

terabytesoftw:
I thought of a method `loadAndValidate($body, $validateInterface)`, what's wrong with `getAttributeCastValue()` being `public`, you may need it.

xepozz:
But there are no use-cases. If you need access to raw data then just don't use forms with validation.
But after validation raw and validated datas are equal.

terabytesoftw:
https://github.com/yiisoft/validator/blob/1d076d62c34281aad601f0a4405dcdc26801d196/src/Validator.php#L48  and how will the validator and widgets access the value if the property is private?

samdark (approved):

PR #186, opened 2022-03-12
Allow nested attributes as array

Gerych1984:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

1. Allow deep nested forms
2. Allow old syntax with **dot** and new separate syntax. For example `getAttributeValue('attribute', 'nested-1', 'nested-2')`
3. Allow both syntax for  **load** nesteds `FormModel`. Previous `attribute.nested = value` and as simple array (for example from $_GET/$_POST)

```
[
    'attribute' => [
        'nested' => 'value',
    ],
]
```
4. Allow nullable property
5. Fix union type error

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -31,6 +31,7 @@ use function substr;
 abstract class FormModel implements FormModelInterface, PostValidationHookInterface, RulesProviderInterface
 {
     private array $attributes;
+    private array $nullable = [];
     private ?FormErrorsInterface $formErrors = null;
     private ?Inflector $inflector = null;
     /** @psalm-var array<string, string|array> */
@@ -47,11 +48,11 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
         return array_keys($this->attributes);
     }
 
-    public function getAttributeHint(string $attribute): string
+    public function getAttributeHint(string $attribute, string ...$nested): string
     {
         $attributeHints = $this->getAttributeHints();
         $hint = $attributeHints[$attribute] ?? '';
-        $nestedAttributeHint = $this->getNestedAttributeValue('getAttributeHint', $attribute);
+        $nestedAttributeHint = (string) $this->getNestedAttributeValue('getAttributeHint', $attribute, ...$nested);
 
         return $nestedAttributeHint !== '' ? $nestedAttributeHint : $hint;
     }
@@ -64,7 +65,7 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
         return [];
     }
 
-    public function getAttributeLabel(string $attribute): string
+    public function getAttributeLabel(string $attribute, string ...$nested): string
     {
         $label = $this->generateAttributeLabel($attribute);
         $labels = $this->getAttributeLabels();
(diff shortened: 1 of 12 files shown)


src/FormModel.php
```diff
@@ -339,14 +340,24 @@ private function writeProperty(string $attribute, $value): void
...
+                    $class->$attribute->setAttribute($nested, $value);
+                } elseif (is_array($value)) {
+                    $class->$attribute->load($value, '');
+                } else {
+                    throw new RuntimeException('$value must be array for using as nested attribute');
+                }
+            } else {
+                $class->$attribute = $value;
+            }
+        };
```
xepozz, line 355:
```suggestion
            $attribute = $class->$attribute;
            if (!is_object($attribute) || !$attribute instanceof self::class) {
                $class->$attribute = $value;
                return;
            }
            if ($nested !== null) {
                $attribute->setAttribute($nested, $value);
                return;
            }
            if (is_array($value)) {
                $attribute->load($value, '');
                return;
            }
            
            throw new RuntimeException('$value must be array for using as nested attribute');
        };
```

tests/FormModelTest.php
```diff
@@ -262,6 +262,22 @@ public function testLoadWithNestedAttribute(): void
...
+        $data = [
+            'FormWithNestedAttribute' => [
+                'user' => [
+                    'login' => 'admin',
+                ],
+            ],
+        ];
+
+        $this->assertTrue($form->load($data));
+        $this->assertSame('admin', $form->getUserLogin());
```
xepozz, line 278:
```suggestion
        $this->assertEquals('admin', $form->getUserLogin());
```

xepozz (changes requested):
Looks good. Do we still need to support notation with `.`?

Gerych1984:
> Looks good. Do we still need to support notation with `.`?

Good day. I use this trick in my project just two days (for generate standart querystrings in js components). I think it can be saved for BC for first time, but also need changes in `HtmlForm` helper for generate input names wo dots. Anyway - now i have two cases, on plain html using `.` from helper, and generated from js as standart arrays

xepozz:
But original HTML Forms send nested attributes with `[]`. So user can use brackets notation instead of dot.

Gerych1984:
> But original HTML Forms send nested attributes with `[]`. So user can use brackets notation instead of dot.

Of course and works with standart arrays more convenient in my opinion. But in current time my most difficult form has max 4th nested levels and input name generated something like `Search[addreesses.k_numbers.fiases.ids]` etc. It only way to generate it using `HtmlForm` helper and it's not was big problem before i need generate same querystring using js/vue.

xepozz:
But what's the difference between `Search[addreesses.k_numbers.fiases.ids]` and `Search[addreesses[k_numbers[fiases[ids]]]]` if it is generated with a class-helper and is parsed by form?

Gerych1984:
> But what's the difference between `Search[addreesses.k_numbers.fiases.ids]` and `Search[addreesses[k_numbers[fiases[ids]]]]` if it is generated with a class-helper and is parsed by form?

Roight format that can be using native php without any parsers must be `Search[addresses][k_numbers][fiases][ids]`. That PR can be first step to do it

terabytesoftw:
Forms must support the `Nested::class` Validator syntax, support .dot format.

PR #191, opened 2022-04-01
More clean-up.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

- [x] Remove @psalm-suppress where possible.
- [x] Change `if{} else{}` for `match()`.
- [x] Fix typehint and phpdoc for `value()` method in `GlobalAttributes.php`.

### DIFF
--- a/src/FormErrors.php
+++ b/src/FormErrors.php
@@ -56,11 +56,10 @@ final class FormErrors implements FormErrorsInterface
 
     public function getFirstError(string $attribute): string
     {
-        if (empty($this->attributesErrors[$attribute])) {
-            return '';
-        }
-
-        return reset($this->attributesErrors[$attribute]);
+        return match (empty($this->attributesErrors[$attribute])) {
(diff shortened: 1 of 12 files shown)


src/FormErrors.php
```diff
@@ -56,11 +56,10 @@ public function getErrorSummaryFirstErrors(): array
 
     public function getFirstError(string $attribute): string
     {
-        if (empty($this->attributesErrors[$attribute])) {
-            return '';
-        }
-
-        return reset($this->attributesErrors[$attribute]);
+        return match (empty($this->attributesErrors[$attribute])) {
```
samdark, line 59: I don't think it is necessary to use match here. It was alright.

terabytesoftw: `Match()` does a strict match plus we only get one return, which is nice.

PR #194, opened 2022-04-08
Remove `Radio` field

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -

Single radio field is unnecessary. Radio buttons always used by group (we have `RadioList` for that).


samdark (approved):
CHANGELOG?

PR #205, opened 2022-06-03
Add methods for set input container class

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Field/Base/PartsField.php
+++ b/src/Field/Base/PartsField.php
@@ -150,6 +150,33 @@ abstract class PartsField extends BaseField
         return $new;
     }
 
+    /**
+     * Replace input container tag CSS classes with a new set of classes.
+     *
+     * @param string|null ...$class One or many CSS classes.
+     */
+    final public function inputContainerClass(?string ...$class): static
+    {
+        $new = clone $this;
+        $new->inputContainerAttributes['class'] = array_filter($class, static fn ($c) => $c !== null);
+        return $new;
+    }
+
+    /**
+     * Add one or more CSS classes to the input container tag.
+     *
+     * @param string|null ...$class One or many CSS classes.
+     */
+    final public function addInputContainerClass(?string ...$class): static
+    {
+        $new = clone $this;
+        Html::addCssClass(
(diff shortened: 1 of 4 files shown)


src/Field/Base/PartsField.php
```diff
@@ -150,6 +150,33 @@ final public function addInputContainerAttributes(array $attributes): static
...
 
+    /**
+     * Replace input container tag CSS classes with a new set of classes.
+     *
+     * @param string|null ...$class One or many CSS classes.
+     */
+    final public function inputContainerClass(?string ...$class): static
+    {
+        $new = clone $this;
+        $new->inputContainerAttributes['class'] = array_filter($class, static fn ($c) => $c !== null);
```
samdark, line 161: Why do we need null check here?

src/Field/Base/PartsField.php
```diff
@@ -150,6 +150,33 @@ final public function addInputContainerAttributes(array $attributes): static
...
+     * Add one or more CSS classes to the input container tag.
+     *
+     * @param string|null ...$class One or many CSS classes.
+     */
+    final public function addInputContainerClass(?string ...$class): static
+    {
+        $new = clone $this;
+        Html::addCssClass(
+            $new->inputContainerAttributes,
+            array_filter($class, static fn ($c) => $c !== null),
```
samdark, line 175: Why do we need null check here?

src/Field/Base/PartsField.php
vjik: For reset class. For example:

```php
// <field class="red blue">
$field = $field->inputContainerClass('red', 'blue');
...
// <field>
$field = $field->inputContainerClass(null);
```

If not filter null values, then result will be `<field class>` instead of `<field>`.

src/Field/Base/PartsField.php
vjik: Agree. Here not need null values in function parameters. I'm remove this in separate PR (at once in all places).

samdark (approved):

PR #207, opened 2022-06-06
Add type hint `array|object|null` to data of `FormModel::load()`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -

**Why?**

Popular case:

```php
$form->load($request->getParsedBody())
```

But `Psr\Http\Message\ServerRequestInterface::getParsedBody()` returns `null|array|object`.

### DIFF
--- a/tests/FormModelTest.php
+++ b/tests/FormModelTest.php
@@ -6,6 +6,7 @@ namespace Yiisoft\Form\Tests;
 
 use InvalidArgumentException;
 use PHPUnit\Framework\TestCase;
+use stdClass;
 use Yiisoft\Form\FormModel;
 use Yiisoft\Form\Tests\TestSupport\CustomFormErrors;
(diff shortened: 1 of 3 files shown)


samdark (approved):

tests/FormModelTest.php
```diff
@@ -261,6 +262,33 @@ public function testLoadWithNestedAttribute(): void
         $this->assertSame('admin', $form->getUserLogin());
     }
 
+    public function testLoadObjectData(): void
```
arogachev, line 265: Maybe use data provider?

arogachev (approved):

tests/FormModelTest.php
samdark: It's alright as is, I think.

arogachev: :ok_hand:

Issue #200, opened 2022-05-23
Yiisoft\Form\Widget\Validator\FieldValidator::getValidatorAttributes() shouldn't apply required attribute for Required rule indiscriminately


hyncica:
### What steps will reproduce the problem?

- Create a form model with 2 attributes.
- Add Required for both attributes with `when` callback set that only one of them needs to be filled.
```php
public function getRules(): array
{
    return [
        'attr1' => [new Required(when: fn() => empty($this->attr2))],
        'attr2' => [new Required(when: fn() => empty($this->attr1)),
    ];
}
```
- Create form with inputs for both attributes.

### What is the expected result?
The form can be send with only one field filled.
If no field is filled the validation won't pass.
 
### What do you get instead?
The form can't be submitted with one field filled because both inputs have `required` attribute.

### Additional info

This is caused by FieldValidator adding `required` for each attribute that has `Required` validation rule indiscriminately.
https://github.com/yiisoft/form/blob/4c4f3fa7a26dca08ad75d479240509df9cf08654/src/Widget/Validator/FieldValidator.php#L40-L42

IMHO the `required` attribute should not be added if the `Required` rule have `when` callback set because it's impossible to predict if the field can or cannot be empty.

Unfortunately the same issue is also present in big-ref-2 branch.
https://github.com/yiisoft/form/blob/28018db50a65f3bfddf32e7ccfe29da2e10c5b38/src/Field/Text.php#L214-L216

| Q                          | A
| ------------------- | ---
| Version                 |  dev-master ([4c4f3fa](https://github.com/yiisoft/form/commit/4c4f3fa7a26dca08ad75d479240509df9cf08654))
| PHP version          | 8.1
| Operating system | Windows 10

PR #209, opened 2022-06-15
Don't enrich fields from rules with non-null "when" parameter

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Fixed issues  | #200


samdark (approved):

samdark:
👍

PR #215, opened 2022-07-22
Allow null with method `setAttribute()` in `FormModel::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/FormModel.php
+++ b/src/FormModel.php
@@ -176,20 +176,22 @@ abstract class FormModel implements FormModelInterface, PostValidationHookInterf
 
     public function setAttribute(string $name, mixed $value): void
     {
-        [$realName] = $this->getNestedAttribute($name);
+        if ($value !== null) {
+            [$realName] = $this->getNestedAttribute($name);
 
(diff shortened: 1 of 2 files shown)


src/FormModel.php
```diff
@@ -176,20 +176,22 @@ public function load(array|object|null $data, ?string $formName = null): bool
...
+                    'bool' => (bool) $value,
+                    'float' => (float) $value,
+                    'int' => (int) $value,
+                    'string' => (string) $value,
+                    default => $value,
+                };
+            }
         }
+
+        $this->writeProperty($name, $value);
```
vjik, line 194: After this change can be try write to non-exist property. Maybe add check on null into `if (isset($this->attributes[$realName])) {` ?

terabytesoftw: done.

vjik (approved):

xepozz (approved):

PR #222, opened 2022-08-03
Ability to handle uploaded files

githubjeka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | 

Code from yii-demo
```php
 public function contact(
        ValidatorInterface $validator,
        ServerRequestInterface $request
    ): ResponseInterface {
        $body = $request->getParsedBody();
        $form = new ContactForm();
        if (($request->getMethod() === Method::POST) && $form->load((array)$body) && $validator
                ->validate($form)
                ->isValid()) {
        $this->mailer->send($form, $request); // <<--- here checking uploaded files :(
        //...
```

Code from Yii2 doc

```php

if (Yii::$app->request->isPost && $model->load(Yii::$app->request->post()) {
            $model->imageFile = UploadedFile::getInstance($model, 'imageFile');
            if ($model->upload()) { // << -- here validate :(
                // file is uploaded successfully
                return;
            }
        }
         //...
```

New approach 

```php

final class FilesForm extends FormModel
{
    public function __construct(
        public ?string $name = null,
        public Files $avatars = new Files([]),
        public Files $images = new Files([]),
        public Files $photos =new Files([])
    ) {
        parent::__construct();
    }
}
```

The most common cases in controller
```php
public function contact(FilesFormRequest $request): ResponseInterface
{
    if ($request->serverRequest()->hasHeader('Content-Type') && $request->isValid()) {
        $this->mailer->send($request->formModel());

        return $this->responseFactory
            ->createResponse()
            ->withHeader(Header::LOCATION, $this->url->generate('site/contact'));
    }

    return $this->viewRenderer->render('form', ['form' => $request->formModel()]);
}
```

Or using handleRequest for upload data and files and using of desired form name

```php
public function contact(
        ValidatorInterface $validator,
        ServerRequestInterface $request
    ): ResponseInterface {
        $form = new ContactForm();        
        if ($request->hasHeader('Content-Type') && $form->handleRequest($request, 'myform') && $validator
                ->validate($form)
                ->isValid()) {
            $this->mailer->send($form, $request);

            return $this->responseFactory
                ->createResponse(Status::OK)
                ->withHeader(Header::LOCATION, $this->url->generate('site/contact'));
        }

        return $this->viewRenderer->render('form', ['form' => $form]);
    }
```

**New features:**
- Files - the collection of upload files. Type of files for using in validations.
- FormModel::handleRequest() - uploads files and data in form model
- FormRequest - for using in most common cases

### DIFF
--- a/src/Files.php
+++ b/src/Files.php
@@ -11,6 +11,9 @@ use IteratorAggregate;
 use Psr\Http\Message\UploadedFileInterface;
 use Yiisoft\Arrays\ArrayAccessTrait;
 
+/**
+ * The collection of upload files
+ */
 final class Files implements Countable, IteratorAggregate, ArrayAccess
 {
     use ArrayAccessTrait;
@@ -32,6 +35,7 @@ final class Files implements Countable, IteratorAggregate, ArrayAccess
                     throw new InvalidArgumentException('Data should contain UploadedFileInterface objects only');
                 }
             }
+            /** @var UploadedFileInterface[] $data */
             $this->data = $data;
         }
     }
--- a/src/Files.php
+++ b/src/Files.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form;
+
+use ArrayAccess;
+use Countable;
+use InvalidArgumentException;
+use IteratorAggregate;
+use Psr\Http\Message\UploadedFileInterface;
+use Yiisoft\Arrays\ArrayAccessTrait;
+
+final class Files implements Countable, IteratorAggregate, ArrayAccess
+{
+    use ArrayAccessTrait;
+
+    /**
+     * @var UploadedFileInterface[]
+     */
+    public array $data;
+
+    public function __construct(mixed $data)
+    {
+        if ($data instanceof UploadedFileInterface) {
+            $this->data = [$data];
+        } elseif (!is_array($data)) {
+            throw new InvalidArgumentException('Data should contain array of uploaded files');
+        } else {
+            foreach ($data as $file) {
+                if (!$file instanceof UploadedFileInterface) {
+                    throw new InvalidArgumentException('Data should contain UploadedFileInterface objects only');
+                }
+            }
+            $this->data = $data;
+        }
+    }
+}
--- a/src/FormRequest.php
+++ b/src/FormRequest.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form;
+
+use Psr\Http\Message\ServerRequestInterface;
+use Yiisoft\Validator\ValidatorInterface;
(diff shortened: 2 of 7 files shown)


src/Files.php
```diff
@@ -0,0 +1,42 @@
...
+        } else {
+            foreach ($data as $file) {
+                if (!$file instanceof UploadedFileInterface) {
+                    throw new InvalidArgumentException('Data should contain UploadedFileInterface objects only');
+                }
+            }
+            /** @var UploadedFileInterface[] $data */
+            $this->data = $data;
+        }
+    }
```
samdark, line 41: Is there a case for a single-file construction?

```suggestion
    /**
     * @param UploadedFileInterface[] $files
     */
    public function __construct(private array $files)
    {
        foreach ($files as $file) {
            if (!$file instanceof UploadedFileInterface) {
                    throw new InvalidArgumentException('Data should contain UploadedFileInterface objects only.');
            }
        }
        
        $this->files = $file;
    }
```

src/Files.php
```diff
@@ -0,0 +1,42 @@
...
+ * The collection of upload files
+ */
+final class Files implements Countable, IteratorAggregate, ArrayAccess
+{
+    use ArrayAccessTrait;
+
+    /**
+     * @var UploadedFileInterface[]
+     */
+    public array $data;
```
samdark, line 24:
```suggestion
```

src/Files.php
samdark: Could use propety promotion. Also, making files private protects it from accidental modification. Should add a getter though.

src/FormRequest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function isValid(): bool
+    {
+        if (!$this->isHandled) {
+            $this->handle();
+        }
+        return !$this->formModel()->getFormErrors()->hasErrors();
+    }
+
+    public function formModel(): FormModel
```
samdark, line 33: The method makes two things at once: handles the form and returns the model. I'd try to avoid it and do these two things separately to avoid side effects.

vjik:
Why need `FormRequest` class? We don't can resolve him from container.
Seems, for use in controller we need helper only that prepare data from request.

```php
public function myAction(ServerRequestInterface $request) {
  $form = new MyForm();
  if (
      $request->getMethod() === Method::POST
      && $form->load(RequestHelper::prepareData($request))
      && $this->validator->validate($form)->isValid()
  ) {
    ...
  }
  ...
```

`RequestHelper::prepareData` should support several strategy (maybe second parameter will be define strategy of data preparing). Need to think about implementation.

vjik:
Use `Yiisoft\Input\Http\Attribute\Parameter\UploadedFiles` isntead of.

PR #229, opened 2022-08-23
Implement `FormCollector::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


tests/FormCollectorTest.php
```diff
@@ -0,0 +1,97 @@
...
+            {
+                return 'joe doe';
+            }
+        };
+
+        $formmodel->setAttribute('string', $objectStringable);
+        $this->assertSame('joe doe', $formmodel->getAttributeValue('string'));
+    }
+
+    public function testSetValue(): void
```
xepozz, line 57: Please use `@dataProvider` instead

src/FormCollector.php
```diff
@@ -0,0 +1,96 @@
...
+use Yiisoft\Validator\RuleInterface;
+
+final class FormCollector
+{
+    private array $attributes;
+    private array $rules = [];
+
+    public function __construct(private FormModelInterface $formModel)
+    {
+        [$this->attributes, $this->rules] = $this->collectorAttributes();
```
xepozz, line 21: Isn't it better to do it in each any public methods to prevent collecting rules and attributes at the creation time? Kind of lazy initialization.

src/FormModel.php
```diff
@@ -113,6 +108,11 @@ public function getAttributeValue(string $attribute): mixed
         return $this->rawData[$attribute] ?? $this->getAttributeCastValue($attribute);
     }
 
+    public function getFormCollector(): FormCollector
```
xepozz, line 111: Is it really needed?

src/FormCollector.php
```diff
@@ -0,0 +1,96 @@
...
+namespace Yiisoft\Form;
+
+use Exception;
+use InvalidArgumentException;
+use ReflectionAttribute;
+use ReflectionNamedType;
+use ReflectionObject;
+use Yiisoft\Validator\RuleInterface;
+
+final class FormCollector
```
xepozz, line 14: I would mark it as `@internal`

xepozz (changes requested):
👍

vjik (changes requested):
Suggest use `ObjectParser` from validator for get rules from form.

PR #244, opened 2023-02-05
Adapt to new widgets feature "theme config"

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -


config/params.php
```diff
@@ -4,9 +4,9 @@
 
 return [
     'yiisoft/form' => [
-        'configs' => [
+        'themeConfigs' => [
```
samdark, line 7: I think these names were alright.

docs/creating-forms.md
```diff
@@ -111,12 +111,12 @@ $this->setTitle($title);
             ->csrf($csrf)
             ->id('form-auth-login')
             ->open() ?>
-            <?= Field::text($formModel, 'login')
+            <?= ThemeDispatcher::text($formModel, 'login')
```
samdark, line 114: That's does not read as good as it was before the change.

tests/ConfigTest.php
```diff
@@ -48,13 +45,13 @@ public function testCustomParams(): void
...
         ];
-        $params['yiisoft/form']['defaultConfig'] = 'simple';
+        $params['yiisoft/form']['defaultThemeName'] = 'simple';
 
         $bootstrapList = $this->getBootstrapList($params);
         $bootstrap = array_shift($bootstrapList);
         $bootstrap(new SimpleContainer());
 
-        $input = Field::hidden(new HiddenForm(), 'key');
+        $input = Hidden::widget()->formAttribute(new HiddenForm(), 'key');
```
samdark, line 54: That is significant regression is readability.

vjik: In master `Field` merge package-specific configurations (common configuration in field factory, individual configuration in field factory, configuration at the place of use) and pass it to widget factory from `yiisoft/widget`.

In this PR use new feature of `yiisoft/widget` (PR)In master `Field` merge package-specific configurations (common configuration in field factory, individual configuration in field factory, configuration at the place of use) and pass to `WidgetFactory` from `yiisoft/widget`.

This PR use new feature of `yiisoft/widget` (see https://github.com/yiisoft/widget/pull/84). Now package-specific configuration merged into widget, `Field` not need for it.

But we can create new `Field` that will be used only as shortcutrs. How is it done in Yii HTML (helper `HTML`).

samdark: Here I'm talking only about how it looks from user perspective. HTML and widgets are used a lot and developer types their code and reads this code constantly so the following is a significant difference:

```php
Field::hidden(new HiddenForm(), 'key');

// vs

Hidden::widget()->formAttribute(new HiddenForm(), 'key');
```

vjik: Yes, I agree with it.

vjik: Goal of PR make `Field::hidden(new HiddenForm(), 'key');` === `Hidden::widget()->formAttribute(new HiddenForm(), 'key');`.

In master `Field::...` use package-specific configuration, but `Hidden::...` don't use. In PR both calls returns same result.

samdark: Then why do we change all tests and phpdoc to use `widget()`?

config/params.php
vjik: revert

docs/creating-forms.md
vjik: fixed

tests/ConfigTest.php
vjik: > Then why do we change all tests and phpdoc to use `widget()`?

Not need. I'm added `Field` with static shortcut methods.

src/Field.php
```diff
@@ -33,238 +32,182 @@
 use Yiisoft\Form\Field\Textarea;
 use Yiisoft\Form\Field\Url;
 
-use function array_key_exists;
-
-final class Field
+class Field
```
samdark, line 35: Is the idea to be able to `extend` it in your app to add more shortcuts?

src/ThemeDispatcher.php
```diff
@@ -0,0 +1,83 @@
...
+
+declare(strict_types=1);
+
+namespace Yiisoft\Form;
+
+use RuntimeException;
+
+use function array_key_exists;
+
+final class ThemeDispatcher
```
samdark, line 11: It doesn't dispatch anything so... maybe a theme factory?

src/Field.php
vjik: Yes. For add shortcuts for non-core fields.

src/ThemeDispatcher.php
vjik: `ThemeContainer` ?

samdark: Better than dispatcher, yes.

vjik: done

samdark (approved):

PR #275, opened 2023-11-22
Allow use fields without model

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | -


src/Field.php
```diff
@@ -55,46 +56,46 @@ final public static function checkbox(
         string $attribute,
         array $config = []
     ): Checkbox {
-        return Checkbox::widget(config: $config)->formAttribute($formModel, $attribute);
+        return Checkbox::widget(config: $config)->inputData(new FormModelInputData($formModel, $attribute));
```
Tigrov, line 59: Could it be loaded from model?
`->inputData($formModel->getInputData($attribute))`
or
`->modelInputData($formModel, $attribute);`

And create `new FormModelInputData()` inside

vjik: > `->inputData($formModel->getInputData($attribute))`

Usage `Field` helper is main way for create widgets, so I didn't add `getInputData()` to form model.

> `->modelInputData($formModel, $attribute);`

It approach add info about form vmodel to fields, but fields do not depend from form model. Fields use `InputDataInterface` implementations only.

Tigrov (approved):

Issue #273, opened 2023-11-18
Allow use fields without form model


vjik:

vjik:
Done by #275 and #281

PR #281, opened 2023-11-25
Add methods `name()` and `value()` for input fields

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #273


src/Field/Base/InputData/FormModelInputData.php
```diff
@@ -102,11 +115,16 @@ public function getPlaceholder(): ?string
...
+     * @return string|null The generated input ID.
      */
-    public function getId(): string
+    public function getId(): ?string
     {
-        $name = mb_strtolower($this->getName(), 'UTF-8');
+        $name = $this->getName();
+        if ($name === null) {
+            return null;
+        }
```
Tigrov, line 125:
```suggestion
        if ($name === '') {
            return '';
        }
```

src/Field/Base/InputData/FormModelInputData.php
```diff
@@ -165,4 +183,20 @@ private function parseProperty(string $property): array
             'suffix' => $matches[3],
         ];
     }
+
+    public function withName(?string $name): static
```
Tigrov, line 187:
```suggestion
    public function withName(string $name): static
```

src/Field/Base/InputData/FormModelInputData.php
```diff
@@ -165,4 +183,20 @@ private function parseProperty(string $property): array
...
+    {
+        $new = clone $this;
+        $new->customName = $name;
+        $new->useCustomName = true;
```
Tigrov, line 191:
```suggestion
```

src/Field/Base/InputData/FormModelInputData.php
```diff
@@ -48,10 +53,14 @@ public function getValidationRules(): iterable
      *
      * @throws InvalidArgumentException If the attribute name contains non-word characters or empty form name for
      * tabular inputs.
-     * @return string The generated input name.
+     * @return string|null The generated input name.
      */
-    public function getName(): string
+    public function getName(): ?string
     {
+        if ($this->useCustomName) {
```
Tigrov, line 60:
```suggestion
        if ($this->customName !== null) {
```

src/Field/Base/InputData/FormModelInputData.php
```diff
@@ -22,6 +22,11 @@ final class FormModelInputData implements InputDataInterface
      */
     private ?iterable $validationRules = null;
 
+    private bool $useCustomName = false;
+    private ?string $customName = null;
```
Tigrov, line 26:
```suggestion
    private ?string $customName = null;
```

Looks like `$useCustomName` is not needed.

Use empty string `''` when set empty `$customName` and check `$customName === null` to get if `$customName` was set.
`->withName('')` instead of `->withName(null)`

The same for `$customValue`

src/Field/Base/InputData/FormModelInputData.php
Tigrov: Review comments is out of order, needs to watch from the end )

src/Field/Base/InputData/FormModelInputData.php
vjik: Empty (`''`) and `null` values are different:
- empty: `<input name>`
- `null`: `<input>`

Also I have redone everything. I removed new methods from `InputDataInterface` and created independent `InputDataWithCustomNameAndValueTrait` trait. I think this is more correct solution.

Tigrov (approved):

PR #317, opened 2024-04-01
Docs

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Fix #7

PR #324, opened 2024-05-18
Fix docs and translate into portuguese

luizcmarin:
In https://github.com/yiisoft/form/pull/317 new docs writing. Wait it.


vjik:
In #317 new docs writing. Wait it.

PR #386, opened 2026-03-18
Bump minimal `yiisoft/html` version to `3.13` and add support of `^4.0`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "php": "8.1 - 8.5",
         "yiisoft/friendly-exception": "^1.0",
-        "yiisoft/html": "^3.8",
+        "yiisoft/html": "^3.13 || ^4.0",
         "yiisoft/widget": "^2.2"
     },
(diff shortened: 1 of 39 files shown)


Tigrov (approved):

composer.json
```diff
@@ -19,7 +19,7 @@
     "require": {
         "php": "8.1 - 8.5",
         "yiisoft/friendly-exception": "^1.0",
-        "yiisoft/html": "^3.8",
+        "yiisoft/html": "^3.13 || ^4.0",
```
Tigrov, line 22: Why need to bump it to `3.13`? Doesn't it work with `3.8`?

vjik: `::tag()` usage is deprecated, but the constructor was made public only in 3.13.

samdark (approved):

PR #391, opened 2026-06-14
Remove deprecated Checkbox::enclosedByLabel()

Mister-42:
## Bugfix

Remove deprecated `Checkbox::enclosedByLabel()` method. Use `labelPlacement()` instead.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️


tests/Field/CheckboxTest.php
vjik, line 1: Check tests. You need to make sure that they don't duplicate the others. Probably, tests with deprecated method can be removed?

Mister-42:
Removed the duplicate tests. Thanks for the review!

src/Field/Checkbox.php
```diff
@@ -334,11 +319,6 @@ private function prepareCheckboxValue(mixed $value): ?string
 
     private function getLabelPlacement(): CheckboxLabelPlacement
```
vjik: `getLabelPlacement` can be removed, and use `$this->labelPlacement` directly.

vjik (changes requested):

vjik (commented):
Seems, removed tests decrease MSI. Kill muatant.

Mister-42:
Updated the data providers instead: DEFAULT yields now don't call `uncheckValue(null)`, keeping the hidden input in the expected output. This covers the code path the removed tests used to cover and should fix the MSI drop.

vjik:
> Updated the data providers instead: DEFAULT yields now don't call `uncheckValue(null)`, keeping the hidden input in the expected output. This covers the code path the removed tests used to cover and should fix the MSI drop.

Now data providers contains too much parameters. Tests looks difficult. Can it be simplified somehow? Maybe instead of edits, add new tests?

vjik (approved):
⚠️ This PR introduces a BC break, so it should be merged after other BC-breaking changes in the repository.

samdark (approved):

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 # Yii Form Change Log
 
-## 1.5.3 under development
+## 2.0.0 under development
 
 - New #376: Allow `ButtonField` instances in `ButtonGroup::buttons()` (@Mister-42)
 - New #388: Add `beforeCheckbox()` and `afterCheckbox()` methods to `Checkbox` and `CheckboxList` fields (@vjik)
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
+- Bug #391: Remove deprecated `Checkbox::enclosedByLabel()` method; use `labelPlacement()` instead (@Mister-42)
```
vjik:
```suggestion
- Chg #391: Remove deprecated `Checkbox::enclosedByLabel()` method; use `labelPlacement()` instead (@Mister-42)
```

Oh, I forgot. We also need to add information to `UPGRADE.md`.

UPGRADE.md
vjik, line 1: See https://yiisoft.github.io/docs/internals/021-changelog-upgrade.html#upgrade

vjik (approved):

PR #392, opened 2026-06-15
Move meta-storm.xml to project root (#381)

Mister-42:
Moves `src/Field/.meta-storm.xml` to project root per PhpStorm convention.


.gitattributes
```diff
@@ -30,6 +30,7 @@
 /.styleci.yml           export-ignore
 /infection.json.dist    export-ignore
 /phpunit.xml.dist       export-ignore
+/.meta-storm.xml        export-ignore
```
vjik: Don't ignore it. `.meta-storm.xml` is required for MetaStorm plugin during work in PhpStorm.

vjik (changes requested):
Add line to changelog

vjik (approved):

PR #399, opened 2026-07-26
Fix #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method

Mister-42:
Adds a `bool $themed` parameter to `ButtonGroup::buttonsData()`.

When `themed: true`, buttons are created via `ButtonField` instances (`Button`, `ResetButton`, `SubmitButton`) instead of raw `Tag\Button` objects. This enables theme support for buttons created from array data. Theme styling only applies when the host application has configured a theme via `ThemeContainer`.

Depends on yiisoft/html#285.

| Q | A |
|---|---|
| Is bugfix? | ❌ |
| New feature? | ✔️ |
| Breaks BC? | ❌ |
| Fixed issues  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
-- Enh #399: Add `$asWidget` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
+- Enh #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
 
 ## 1.5.2 March 20, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
+- Enh #399: Add `ButtonGroup::buttonsFieldData()` method (@va108)
 
 ## 1.5.2 March 20, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
-- Enh #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
+- New #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
 
 ## 1.5.2 March 20, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
-- Enh #399: Add `ButtonGroup::buttonsFieldData()` method (@va108)
+- Enh #399: Add `ButtonGroup::buttonsFieldData()` method (@Mister-42)
 
 ## 1.5.2 March 20, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
-- Enh #399: Add `ButtonGroup::buttonsFieldData()` method (@Mister-42)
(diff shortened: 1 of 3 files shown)


src/Field/ButtonGroup.php
```diff
@@ -56,6 +58,57 @@ public function buttonsData(array $data, bool $encode = true): self
...
+    {
+        $buttons = [];
+        foreach ($data as $item) {
+            $label = $item[0] ?? null;
+            $attributes = array_slice($item, 1, null, true);
+
+            $type = $attributes['type'] ?? 'button';
+            unset($attributes['type']);
+
+            $buttonField = match ($type) {
```
vjik: This is not a form field — it's a widget. It seems we should move it to `Yiisoft\Html\Widget\ButtonGroup`.

Here's my vision:

1. Add a third parameter to the existing `buttonsData()` method, e.g. `bool $parseAsWidget = false` (a better name for this parameter would be good). When the new parameter is `true`, use the logic proposed in this PR.

2. Document in the PHPDoc that the default value of this parameter will change to `true` in the next major version.

3. After a minor release of Yii HTML, add a similar parameter to Yii Form.

Mister-42: I have created yiisoft/html#285 and adjusted this PR

Is that in the direction of what you meant?

vjik: Oh, sorry. I thought these were widgets from Yii HTML. Can you remove changes in Yii HTML and adopt current code to it?

Approach with new parameter in `buttonsData()` looks good.

Mister-42: How about this?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - New #388: Add `beforeRadio()` and `afterRadio()` methods to `RadioList` field (@vjik)
 - Enh #386: Bump minimal `yiisoft/html` version to `^4.2` (@vjik)
 - Enh #392: Move `.meta-storm.xml` to the project root (@Mister-42)
+- Enh #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
```
vjik:
```suggestion
- New #399: Add `$themed` parameter to `ButtonGroup::buttonsData()` method (@Mister-42)
```

src/Field/ButtonGroup.php
```diff
@@ -48,11 +51,54 @@ public function buttons(ButtonTag|ButtonField ...$buttons): self
...
+            return $new;
+        }
+
+        $buttonTags = [];
+        foreach ($data as $item) {
+            if (is_array($item)) {
+                $label = $item[0] ?? null;
+                $attributes = array_slice($item, 1, null, true);
+
+                $type = $attributes['type'] ?? 'button';
```
vjik: Make `$type` case insensitive.

vjik (changes requested):

src/Field/ButtonGroup.php
vjik: Yes! Looks good.

src/Field/ButtonGroup.php
Mister-42: Change applied

tests/Field/ButtonGroupTest.php
```diff
@@ -297,12 +297,231 @@ public function testButtonsWithFieldButtonCustomized(): void
...
+                        ],
+                    ],
+                ],
+            ],
+            'default',
+        );
+
+        $result = Field::ButtonGroup()
+            ->buttonsData([
+                [null, 'type' => 'reset'],
```
vjik, line 451: Add test w/o label  (`['type' => 'button']`). This should be equals to `[null, 'type' => 'button']`

src/Field/ButtonGroup.php
```diff
@@ -6,9 +6,13 @@
 
 use Yiisoft\Form\Field\Base\ButtonField;
 use Yiisoft\Form\Field\Base\PartsField;
+use Yiisoft\Form\Theme\ThemeContainer;
 use Yiisoft\Html\Tag\Button as ButtonTag;
 use Yiisoft\Html\Widget\ButtonGroup as ButtonGroupWidget;
```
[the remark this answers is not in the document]
Mister-42: That is incorrect. cs-fixer does see that as used and does not fail

tests/Field/ButtonGroupTest.php
Mister-42: `[]` should even be equal to `[null, 'type' => 'button']`

src/Field/ButtonGroup.php
```diff
@@ -6,9 +6,13 @@
...
 
+use function array_slice;
+use function is_array;
+
```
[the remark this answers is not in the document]
Mister-42: Wasn't reported for some reason

vjik (approved):

src/Field/ButtonGroup.php
```diff
@@ -48,11 +53,54 @@ public function buttons(ButtonTag|ButtonField ...$buttons): self
...
+                if ($label !== null) {
+                    $buttonWidget = $buttonWidget->content((string) $label);
+                }
+
+                $buttonTags[] = $buttonWidget
+                    ->encodeContent(false)
+                    ->addButtonAttributes($attributes)
+                    ->getButton()
+                    ->encode($encode);
+            }
```
samdark, line 84: Validation behavior with `themed: true` is different:

```php
ButtonGroup::widget()
      ->buttonsData([
          'invalid',
          ['Send', 'type' => 'submit'],
      ], themed: true)
      ->render();
```

Will silently skip `invalid` but when `themed: false`, `InvalidArgumentException` will be thrown. There's `testButtonsDataThemedSkipsNonArrayItems` for that but I don't think that this behavior change is correct.

Should be consistent with current behavior.

samdark (changes requested):

src/Field/ButtonGroup.php
Mister-42: Fixed in 9b29ed0 — non-array items now throw `InvalidArgumentException` in the themed path too, consistent with the non-themed behavior. Updated `testButtonsDataThemedSkipsNonArrayItems` accordingly (renamed to `testButtonsDataThemedThrowsExceptionForNonArrayItem`).

samdark (approved):

